PR #1, opened 2019-04-16
Absorb go-libp2p abstractions and core types into this module

raulk:
This PR is part of a batch across go-libp2p repos to consolidate abstractions, interfaces and core types into this repo.

Work is being performed in the `consolidate-skeleton` branch across repos.

---

Master plan:

- [X] Iteration 1: focused on copying over the abstractions with verbatim godoc and making no breaking changes.
- [X] Iteration 2: revising godoc to make it top-class // proposed owner: @yusefnapora with @raulk collaborating.
- [ ] Iteration 3: minor, aesthetic backwards-incompatible changes, e.g. Ttl => TTL. Manifest (keep adding):
  * https://github.com/libp2p/go-libp2p-core/pull/1#discussion_r276047782
  * https://github.com/libp2p/go-libp2p-core/pull/1#discussion_r277135595
  * https://github.com/libp2p/go-libp2p-core/pull/1#discussion_r277135702

---

Merge strategy: squash and rebase.


connmgr/null.go
```diff
@@ -0,0 +1,20 @@
+package connmgr
+
+import (
+	"context"
+
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+type NullConnMgr struct{}
```
lanzafame: @raulk Out of curiosity, what is the purpose of the null implementation?

protocol/switch.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"io"
+)
+
+// HandlerFunc is a user-provided function used by the Router to
+// handle a protocol/stream.
+type HandlerFunc = func(protocol string, rwc io.ReadWriteCloser) error
+
+type Router interface {
```
lanzafame: Please add a comment to interfaces in this file.

routing/notifications/query.go
```diff
@@ -0,0 +1,120 @@
...
+	if ich == nil {
+		return
+	}
+
+	// We *want* to panic here.
+	ech := ich.(*eventChannel)
+	ech.send(ctx, ev)
+}
+
+func (qe *QueryEvent) MarshalJSON() ([]byte, error) {
```
lanzafame: I have no preference for or against but could we either be consistent with the `*_serde.go` pattern or merge the serde functions into the main type definition files?

routing/routing.go
```diff
@@ -0,0 +1,123 @@
...
+	//
+	// Useful when you want a result *now* but still want to hear about
+	// better/newer results.
+	//
+	// Implementations of this methods won't return ErrNotFound. When a value
+	// couldn't be found, the channel will get closed without passing any results
+	SearchValue(context.Context, string, ...options.Option) (<-chan []byte, error)
+}
+
+// IpfsRouting is the combination of different routing types that IPFS uses.
```
lanzafame:
```suggestion
// Routing is a combination of different routing types.
```

routing/routing.go
```diff
@@ -0,0 +1,123 @@
...
+	// Boostrapped state and remain there. It is not a synchronous call.
+	Bootstrap(context.Context) error
+
+	// TODO expose io.Closer or plain-old Close error
+}
+
+// PubKeyFetcher is an interfaces that should be implemented by value stores
+// that can optimize retrieval of public keys.
+//
+// TODO(steb): Consider removing, see #22.
```
lanzafame:
```suggestion
// TODO(steb): Consider removing, see https://github.com/libp2p/go-libp2p-routing/issues/22.
```

transport/transport.go
```diff
@@ -0,0 +1,95 @@
...
+// method
+type Listener interface {
+	Accept() (UpgradedConn, error)
+	Close() error
+	Addr() net.Addr
+	Multiaddr() ma.Multiaddr
+}
+
+// Network is an inet.Network with methods for managing transports.
+type TransportBasedNetwork interface {
```
lanzafame:
```suggestion
type TransportNetwork interface {
```
Which ever name is chosen, the comment needs updating as well

peer/info.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+	"strings"
+
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// PeerInfo is a small struct used to pass around a peer with
+// a set of addresses (and later, keys?).
+type Info struct {
```
vyzo: Please don't rename the type, it should still be called `PeerInfo`.

lanzafame: Except it now stutters with the package name or always did

connmgr/null.go
raulk: It's used by the blank host: https://github.com/libp2p/go-libp2p-blankhost/blob/master/blank.go#L30.

discovery/discovery.go
```diff
@@ -0,0 +1,48 @@
...
+	Advertiser
+	Discoverer
+}
+
+// DiscoveryOpt is a single discovery option.
+type DiscoveryOpt func(opts *DiscoveryOpts) error
+
+// DiscoveryOpts is a set of discovery options.
+type DiscoveryOpts struct {
+	Ttl   time.Duration
```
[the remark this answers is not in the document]
raulk: I'm favourable, but this is a backwards-incompatible change. I propose to postpone it to a later iteration, before we release go-libp2p-core to GA.

host/host.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/libp2p/go-libp2p-core/connmgr"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/peerstore"
+	"github.com/libp2p/go-libp2p-core/protocol"
+
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// Host is an object participating in a p2p network, which
```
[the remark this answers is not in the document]
raulk: I copied over the original godoc. They needed to be revised heavily anyway, everywhere. Doing so is much approachable now that all abstractions live under a single roof. I'll own that, along with @yusefnapora. But will probably save it for subsequent iteration.

metrics/interface.go
```diff
@@ -0,0 +1,16 @@
+package metrics
```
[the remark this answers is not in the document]
raulk: Done.

discovery/discovery.go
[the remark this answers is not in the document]
vyzo: Can we move over the actual options as well? There is no point in having them in the discovery module, which now houses implementations.

network/notifiees.go
```diff
@@ -0,0 +1,77 @@
...
+func (nb *NotifyBundle) ClosedStream(n Network, s Stream) {
+	if nb.ClosedStreamF != nil {
+		nb.ClosedStreamF(n, s)
+	}
+}
+
+// Global noop notifiee. Do not change.
+var GlobalNoopNotifiee = &NoopNotifiee{}
+
+type NoopNotifiee struct{}
```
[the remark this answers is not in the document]
raulk: I thought of that, but there is a nuisance, and it's totally subjective.

IMO `Noop` makes a strong concrete assertion that this will literally do nothing when called. `Null` is less strict: it conveys that _functionally speaking_ this is equivalent to having no connection manager. Underneath it may not noop, or it may actually do something, but its behaviour as externally observed is akin to a null object.

Granted, this is one man's opinion. And the benefit in DX of unifying naming may trump pragmatism. No strong opinions, but if we are to unify, I'd pick Noop.

peer/info.go
```diff
@@ -0,0 +1,80 @@
...
+
+var _ fmt.Stringer = Info{}
+
+func (pi Info) String() string {
+	return fmt.Sprintf("{%v: %v}", pi.ID, pi.Addrs)
+}
+
+var ErrInvalidAddr = fmt.Errorf("invalid p2p multiaddr")
+
+func InfoFromP2pAddr(m ma.Multiaddr) (*Info, error) {
```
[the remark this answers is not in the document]
raulk: Postponed for Iteration 2 (godoc).

peer/info.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return &Info{
+		ID:    id,
+		Addrs: addrs,
+	}, nil
+}
+
+func InfoToP2pAddrs(pi *Info) ([]ma.Multiaddr, error) {
+	var addrs []ma.Multiaddr
+	tpl := "/" + ma.ProtocolWithCode(ma.P_IPFS).Name + "/"
```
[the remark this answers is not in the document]
raulk: SGTM.

peer/set/peerset.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"sync"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// PeerSet is a threadsafe set of peers
+type PeerSet struct {
+	ps   map[peer.ID]struct{}
+	lk   sync.RWMutex
```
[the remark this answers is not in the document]
raulk: This was copied verbatim, but no reason not to improve it.

routing/notifications/query.go
raulk: Good idea.

protocol/switch.go
raulk: Iteration 2 cc @yusefnapora.

routing/routing.go
raulk: Good catch.

transport/transport.go
raulk: `TransportBasedNetwork` _is_ a mouthful, yet `TransportNetwork` lacks definition. But I take it.

raulk:
@vyzo 

> Can we call the network directory/package just net?

I had initially called it `net` but it overshadows stdlib `net`, and it gets disconcerting.

peer/info.go
raulk: `peer.PeerInfo` is not idiomatic. It used to be `pstore.PeerInfo`, that's why the `Peer` qualifier was required. This type is now in the place where it belongs, and the qualifier becomes redundant.

However, `peer.Info` _is_ vague and unhelpful. Possible alternatives:

* `peer.DialInfo`
* `peer.AddrInfo`
* `peer.AddressSlip`

vyzo: `peer.AddrInfo` seems the most pertinent; it _is_ addr information about a peer.

discovery/discovery.go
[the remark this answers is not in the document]
raulk: Done in 4e1f238839bd0cfa806ca2cc129fcd2bbc2aa0bf.

host/host.go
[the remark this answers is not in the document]
raulk: Iteration 2.

network/network.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/libp2p/go-libp2p-core/protocol"
+
+	ic "github.com/libp2p/go-libp2p-crypto"
+
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// MessageSizeMax is a soft (recommended) maximum for network messages.
+// One can write more, as the interface is a stream. But it is useful
+// to bunch it up into multiple read/writes when the whole message is
```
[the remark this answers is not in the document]
raulk: Iteration 2.

routing/routing.go
raulk: Iteration 2.

network/notifiees.go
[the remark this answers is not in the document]
lanzafame: Yeah, I am fine with Noop, just want to take this chance to unify on something.

transport/test/utils.go
```diff
@@ -0,0 +1,45 @@
...
+	"runtime"
+	"testing"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/transport"
+
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+var Subtests = []func(t *testing.T, ta, tb transport.Transport, maddr ma.Multiaddr, peerA peer.ID){
```
lanzafame: Iteration 2: documentation on how these subtests style tests can be used.

lanzafame: Iteration 2: documentation on how these subtests style tests can be used.

helpers/stream.go
```diff
@@ -0,0 +1,56 @@
+package helpers
```
Stebalien: OT: these should _really_ go away. We need to fix the stream interfaces and:

1. Make `Close()` close.
2. Add `CloseWrite()`, `CloseRead()`.

metrics/meter.go
```diff
@@ -0,0 +1,44 @@
+package metrics
```
Stebalien: Let's not move this library, it's not actually libp2p specific and people should be able to use it outside of libp2p. Really, I only put it in the libp2p org because I didn't want to leave it as a personal repo.

helpers/stream.go
raulk: Since both funcs take a stream as their first arg, I wonder if we can preserve backwards compatibility by hoisting the instance method to a package var. Will try it out, otherwise I suggest punting it to iteration 3 (backwards-incompatible ergonomics changes).

metrics/meter.go
raulk: The metrics interfaces reference libp2p types (peer.ID and protocol.ID). They aren’t used beyond libp2p or projects that transitively depend on libp2p: https://godoc.org/github.com/libp2p/go-libp2p-metrics?importers. I think it’s pretty tied to libp2p.

helpers/stream.go
Stebalien: I agree.

metrics/meter.go
Stebalien: I'm talking about the files taken from https://github.com/libp2p/go-flow-metrics.

raulk: Ah, I see. I did check the importers of go-flow-metrics to determine whether to pull it in or not. [The grand total is 1: us](https://godoc.org/github.com/libp2p/go-flow-metrics?importers).

The library is not actively maintained (no significant commits since December 2017), so I don't think we'll be doing any damage if we pulled it into core.

raulk:
@warchant @kamilsa – this is the PR I'd like you to poke about. Disregard the godocs for now (they'll improve dramatically in iteration 2, see description). But it'd be super helpful to have fresh pairs of eyes go through the abstraction tree and see if it's easily intelligible and intuitive, or if there's room to rename entities or shuffle things around for better clarity.

helpers/stream.go
vyzo: We might call the `FullyClose` as `Shutdown`.

metrics/meter.go
vyzo: I don't think this library belongs to core either; let's not bloat.

vyzo: Besides, aren't we phasing it out in favour of opencensus?

mux/mux.go
```diff
@@ -0,0 +1,56 @@
...
+var ErrReset = errors.New("stream reset")
+
+// Stream is a bidirectional io pipe within a connection.
+type MuxedStream interface {
+	io.Reader
+	io.Writer
+
+	// Close closes the stream for writing. Reading will still work (that
+	// is, the remote side can still write).
+	io.Closer
```
Warchant: here it should be reflected in the method name that it closes stream for writing

mux/mux.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// IsClosed returns whether a connection is fully closed, so it can
+	// be garbage collected.
+	IsClosed() bool
+
+	// OpenStream creates a new stream.
+	OpenStream() (MuxedStream, error)
+
+	// AcceptStream accepts a stream opened by the other side.
+	AcceptStream() (MuxedStream, error)
```
Warchant: Maybe ListenForStreams? (if we name "server" as Listener)

mux/test/muxer.go
```diff
@@ -0,0 +1,541 @@
+package test
```
Warchant: I have a feeling that having `muxer_test.go` is better, rather than having separate test folder

network/conn.go
```diff
@@ -0,0 +1,58 @@
...
+	io.Closer
+
+	ConnSecurity
+	ConnMultiaddrs
+
+	// NewStream constructs a new Stream over this conn.
+	NewStream() (Stream, error)
+
+	// GetStreams returns all open streams over this conn.
+	GetStreams() []Stream
```
Warchant: does Conn work with streams? I thought only MuxedConn does

sec/test/ttest.go
```diff
@@ -0,0 +1,373 @@
+package test
```
Warchant: ttest.go?

mux/mux.go
vyzo: that's a breaking change and we'd like to avoid those.

mux/mux.go
vyzo: again, we are not renaming interface methods here, it's a breaking change.

mux/test/muxer.go
raulk: These are conformance tests _for muxers_, rather than tests _of muxers_. The mux adaptors call out to these to test themselves, e.g. https://github.com/whyrusleeping/go-smux-yamux/blob/master/yamux_test.go

mux/mux.go
raulk: Good point, but yeah, we can't rename _methods_ as that would create breakage, but we can rename _types_. We could consider this for iteration 3 (small breakages).

network/conn.go
raulk: These methods do indeed seem a bit redundant. Adding it to iteration 3 for potential refactor.

mux/mux.go
raulk: Added it to the manifest for iteration 3 (in the PR description).

mux/mux.go
raulk: It's also convention in Go to use Listen to set up a listener service, and Accept to consume incoming events from that listener service, analogous to listen and accept in POSIX.

sec/test/ttest.go
raulk: Good call. Should be transport.go.

raulk:
How would we feel about moving all conformance tests and test helpers to a `go-libp2p-testing` package?

metrics/meter.go
Stebalien: > Ah, I see. I did check the importers of go-flow-metrics to determine whether to pull it in or not. The grand total is 1: us.

I wrote it for us and haven't advertised it so that's not exactly surprising. However, if we pull it into core, nobody outside of libp2p will be able to use it without also pulling in libp2p.

Please _do not_ include this in the core bundle.

> The library is not actively maintained (no significant commits since December 2017), so I don't think we'll be doing any damage if we pulled it into core.

Three haven't been any significant commits since December 2017 because there haven't been any bugs since December 2017. It's completely finished and actively maintained by me (should someone find a bug).

Stebalien: @vyzo go-libp2p-metrics and go-flow-metrics are two different libraries. We use this library for bandwidth metrics.

Stebalien:
> How would we feel about moving all conformance tests and test helpers to a go-libp2p-testing module?

Sounds like a good idea.

metrics/meter.go
raulk: @Stebalien gotcha, will revert `go-flow-metrics`. Sidenote: I re-read the conversation and I think I may have come off a bit abrupt. Apologies, it wasn't my intention.

raulk: Revisiting this code to back out the change, I realise why I decided to coalesce both modules. go-libp2p-metrics _depends on_ go-flow-metrics. The `BandwidthCounter` type uses `flow.Meter` to reify the bookkeeping.

We have two options:

1. pull `Reporter` and its associated struct `Stats` into go-libp2p-core, leaving `BandwidthCounter` behind (in go-libp2p-metrics), considering it the _glue_ between the abstraction and the implementation using go-flow-metrics.
  * I dislike this option because we reduce module count by zero (concerning metrics), and the resulting modules are very slim by themselves.
2. pull all of go-libp2p-metrics in, and make all of go-libp2p-core depend on go-flow-metrics.

I'm choosing option 2 because we're moving to OC anyway. Shout if you disagree.

Stebalien: > Sidenote: I re-read the conversation and I think I may have come off a bit abrupt. Apologies, it wasn't my intention.

Not at all, I was just afraid I wasn't making my position on this clear. I consider go-flow-metrics to be something like go-buffer-pool, useful outside of libp2p, so I really didn't want to see it absorbed.

> pull all of go-libp2p-metrics in, and make all of go-libp2p-core depend on go-flow-metrics.

SGTM. go-libp2p-metrics is very much a libp2p package. BandwidthCounter would, ideally, go in a separate package (maybe even in go-libp2p?) as it's an implementation of the `Reporter` interface but I don't feel strongly either way.

(I'm also not really a fan of the interface itself but we can revisit that in a future refactor)

Stebalien:
Remaining question: global variables.

Unfortunately, I'm pretty sure our only option is to break things. The alternative is to:

1. Leave implementations in the sub-packages.
2. Put the aliases in _this_ package.

But that's even worse.

In general, we can limit the fallout by replacing variables with constants. That way, broken code simply doesn't compile.

So, type by type:

**Errors**

Copy them. Easy.

**Global Variables**

Flags, Timeouts, counts, etc. Ideally, we'd turn as many as we can into constants. However, there are two edge-cases:

First, testing. In some cases, we can deal with this by defining a public constant and a private variable:

```go
const MyTimeout = ...
var myTimeout = MyTimeout

// use `myTimeout` internally.
```

However, this could be a bit tricky if we move tests out of the package. The _ideal_ fix would be to make all these variables configurable at a per-instance/node level.

Second, flags: We may just have to break these. We do try to avoid these for anything but testing so that shouldn't be _that_ much of an issue.

**Registries**

These are true global variables that actually need to be _variables_. As far as I can tell, this is limited to the crypto package ({Pub,Priv}KeyUnmarshallers).

Solution:

1. We should probably make these private.
2. We should add public methods for registering new key types.

transport/transport.go
```diff
@@ -0,0 +1,95 @@
...
+var DialTimeout = 60 * time.Second
+
+// AcceptTimeout is the maximum duration an Accept is allowed to take.
+// This includes the time between accepting the raw network connection,
+// protocol selection as well as the handshake, if applicable.
+var AcceptTimeout = 60 * time.Second
+
+// UpgradedConn is a connection that is is an extension of the net.Conn interface that provides multiaddr
+// information, and an accessor for the transport used to create the conn
+type UpgradedConn interface {
```
Stebalien: This may not have gone through the "upgrader" and really shouldn't be called `UpgradedConn`. Any reason not to just call it a `TransportConn` or a `Conn`?

We'll have to rethink this anyways with the switch refactor but I don't want to imply any mandatory relationship with the upgrader.

yusefnapora: I found this confusing, fwiw. 

Before I really started really reading the go code, I was thinking of transport connections as more or less "raw" network connections and thought that the upgrader would upgrade *from* transport connections to secure + multiplexed connections. 

Looking at the implementations it seems that you pass an upgrader into the transport constructor, and it's the transport's job to apply it to the raw connections. That's fine, but the relationship between transports and the upgrader isn't clear from reading the Transport interface.

raulk: In my head, I take the term "transport" mean things like TCP, UDP, WSS, WebRTC, QUIC, etc, so `TransportConn` wouldn't necessary denote that the connection has the capabilities required by libp2p, I'd take it to represent a raw tarnsport connection.

What I want to convey is that the libp2p stack _expects_ _**this**_ connection to support multiplexing, and to be encrypted, whether native or shimmed.

Stebalien: I agree this is all really confusing. I just find calling this `UpgradedConn` even more confusing.

Really, It doesn't matter _that_ much as we'll need to revisit this in the switch refactor anyways.

kamilsa: Just some thoughts. In C++ implementation we try to keep transport without any dependencies on Upgrader or UpgradedConnections. To this end we want to create a separate folder for ConnUpgrader and everything related to the upgrader (including UpgradedConn if we have one in future). You might consider this approach as well

raulk: I'm also having second thoughts on calling this `UpgradedConn`.

@kamilsa – we also don't have a relationship between the connection and the upgrader, so this naming is definitely misleading if it made you think so. Thanks for the input. Note that the upgrader is a convenience component – it has no associated abstraction, but definitely plays a key role in shimming multiplexing and security in transports that do not support it natively.

Alternative names, off the top of my head:
* `CapableConn`: denotes that it has the capabilities libp2p requires.
* `P2PConn`

Stebalien:
> Thanks to go's aversion to package import cycles, moving test helpers to go-libp2p-testing doesn't work. Tests in go-libp2p-core would use the helpers in go-libp2p-testing, which in turn import go-libp2p-core. Kaput.

I'd either:

1. Leave all tests in go-libp2p-core.
2. Move all tests (not just tools) out of go-libp2p-core.

raulk:
@Stebalien – good news, I managed to break the import cycle by placing the peer and crypto unit tests in separate `_test` packages. See:

https://github.com/libp2p/go-libp2p-core/pull/1/commits/757c750aa1ddd540b64075b010ccad5de688d60a

https://github.com/libp2p/go-libp2p-testing/pull/1/commits/1a0da3de1915915f56eb4c73ca07751b3f74bcb2

Looking good.

Stebalien:
> Therefore, I've migrated only conformance tests, and the one-off test helper (race), but I'm having second thoughts about this separation.

What are your second thoughts?

raulk:
@stebalien

> What are your second thoughts?

Not applicable any longer, now that I managed to snap off the peer and crypto test utils into go-libp2p-testing. I think this separation is the way to go forward, as the go toolchain should not even download go-libp2p-testing for users depending on runtime code. That module becomes part of the SDK, used only by module developers.

Stebalien:
@raulk there are some serious causality violations in this thread... Did you repost that comment or is github having problems?

raulk:
@stebalien haha the second! 

https://twitter.com/raulvk/status/1126178378304970752?s=21

https://mobile.twitter.com/search?q=GitHub%20future

This is the one that’s out of order: https://github.com/libp2p/go-libp2p-core/pull/1#issuecomment-490491201

transport/transport.go
Stebalien: More on the naming bikeshed:

* `Libp2pTransportConn`
* `P2PTransportConn`
* Just leave it as `transport.Conn`?

Regardless, `transport.Listener` should match (e.g, `UpgradedListener`).

raulk: We had consensus around `CapableConn` -- I just renamed it. @raulk, @yusefnapora, @Stebalien signed off on it. @vyzo had some reservations, but I think it'll grow on people. If after a reasonable period people are uncomfortable, we can rename and alias it.

raulk:
@yusefnapora thanks for owning the godocs review and update; great job! I've merged it onto this branch, so the repo will be inited with iteration 1 and iteration 2 completed. This is awesome news!

helpers/stream.go
```diff
@@ -0,0 +1,56 @@
...
+	"errors"
+	"io"
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/network"
+)
+
+// EOFTimeout is the maximum amount of time to wait to successfully observe an
+// EOF on the stream. Defaults to 60 seconds.
+var EOFTimeout = time.Second * 60
```
vyzo: this will break backwards compatibility, the variable needs to be set in `core` and can't be set at the leaf.

peerstore/peerstore.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// PeerInfo returns a peer.PeerInfo struct for given peer.ID.
+	// This is a small slice of the information Peerstore has on
+	// that peer, useful to other services.
+	PeerInfo(peer.ID) peer.AddrInfo
+
+	GetProtocols(peer.ID) ([]string, error)
+	AddProtocols(peer.ID, ...string) error
+	SetProtocols(peer.ID, ...string) error
+	SupportsProtocols(peer.ID, ...string) ([]string, error)
```
vyzo: we should make the protocol stuff into its own interface and compose, I am working to split that off in the peerstore.

Warchant: Host.Mux() may have those methods, since it knows about all protocols (via setProtocolHandler)

helpers/stream.go
vyzo: I recommend removing all such configuration variables from the leaves.
Better to break the build than have an assignment that has no effect.

Stebalien: We can make them constants.

vyzo: Can we make them constants in the leaves while being initialized by a variable?
If that's the case, then it would be fine as it would break assignments.
But we really don't want to turn them into constants in the core module, they are variable for good reason.

Stebalien: Hm. No, we can't. I mean, ideally we'd have some kind of global config...

vyzo:
the peerstore interface has changed too -- there is a split of the protocol methods in the `ProtoBook` interface.

protocol/switch.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// Upon receiving an incoming stream request, the Router will check all registered
+// protocol handlers to determine which (if any) is capable of handling the stream.
+// The handlers are checked in order of registration; if multiple handlers are
+// eligible, only the first to be registered will be invoked.
+type Router interface {
+
+	// AddHandler registers the given handler to be invoked for
+	// an exact literal match of the given protocol ID string.
+	AddHandler(protocol string, handler HandlerFunc)
```
Warchant: why "Add" and not "Set"?
Add says to me that single protocol can be processed by multiple handlers
Set says to me that single protocol can be processed with exactly one handler

raulk: Yeah this is bad naming, indeed. Could you file a separate issue in this repo so we can consider a rename?

raulk:
I'm not going to address the global var situation as there is no clean way of solving it. We can choose to drop the vars in the donor packages, but that'll break consumers. It's probably worth it, though, much better than a silent mess.

helpers/stream.go
raulk: I think we should intentionally break the donor modules by removing those vars at origin. Much better than a silent change in behaviour.

vyzo:
Yeah, let's drop them from the source packages, better break the build than a bug.

helpers/stream.go
raulk: I'll open an issue for this.

peer/set.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"sync"
+)
+
+// PeerSet is a threadsafe set of peers
+type Set struct {
+	lk sync.RWMutex
+	ps map[ID]struct{}
+
+	size int
```
vyzo: the size field seems unused.

raulk: The field is used internally in `TryAdd()`.

PR #4, opened 2019-05-14
Add common OpenCensus metrics registration functions to libp2p-core

mvid:
As we expected multiple products/libraries to begin taking advantage of metrics gathering, we want a central location that allows for registration/lookup of OpenCensus views, rather than duplicating that logic across all of our distributed repositories


metrics/register.go
```diff
@@ -0,0 +1,24 @@
...
+func RegisterViews(name string, views []view.View) {
+	views, ok := registeredViews[name]
+	if ok {
+		panic(fmt.Errorf("duplicate registration of views by name %s", name))
+	}
+
+	registeredViews[name] = views
+}
+
+func LookupViews(name string) []view.View {
```
bigs: let's have this potentially return error if the module hasn't registered

metrics/register.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+
+	"go.opencensus.io/stats/view"
+)
+
+var registeredViews = map[string][]view.View{}
+
+func RegisterViews(name string, views []view.View) {
```
bigs: likewise let's return an error rather than panicking here. i've thought it over and i think that's best!

bigs:
roping in @raulk when he has a moment

.gitignore
```diff
@@ -0,0 +1,2 @@
+# jetbrains
+.idea/
```
raulk: Could you add this to your global gitignore? We don't want proliferation of gitignore files everywhere for different IDEs and all.

metrics/register.go
```diff
@@ -0,0 +1,38 @@
+package metrics
+
+import (
+	"fmt"
+
+	"go.opencensus.io/stats/view"
+)
+
+var registeredViews = map[string][]*view.View{}
```
raulk: This map should be made threadsafe via a mutex.

metrics/register.go
```diff
@@ -0,0 +1,38 @@
...
+		return nil, fmt.Errorf("no view %s found registered", name)
+	}
+	return views, nil
+}
+
+func AllViews() (views []*view.View) {
+	for _, v := range registeredViews {
+		views = append(views, v...)
+	}
+	return
```
raulk: Let's make the return explicit.

```suggestion
	return views
```

metrics/register.go
raulk: Also document this type; it's not obvious that the string is actually a namespace.

metrics/register.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+
+	"go.opencensus.io/stats/view"
+)
+
+var registeredViews = map[string][]*view.View{}
+
+func RegisterViews(name string, views []*view.View) error {
```
raulk: Godocs missing everywhere.

raulk:
@mvid please document the change and the motivation in the title and the description of the PR.

.gitignore
mvid: Ok. Should I remove it from [go-libp2p](https://github.com/libp2p/go-libp2p/commit/ce95b37cd31edb3d9e4a4458cb43be0c9c77704f#diff-a084b794bc0759e7a6b77810e01874f2) as well?

raulk: no need

metrics/register.go
mvid: Updated the documentation to better describe what the map is storing. Also changed it to `sync.Map{}`, is that acceptable?

metrics/register.go
mvid: Added godocs, let me know if they are not clear enough

metrics/register.go
raulk: A Mutex is the standard go bread and butter for sync. Did you need something specific from sync.Map?

mvid: Not sure. It seemed more appropriate, because of this line ([from godocs](https://golang.org/pkg/sync/#Map)): 
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow

Seeing as how this is a write only map, it seemed to make sense. If you prefer the standard map for type safety, I can write it back.

mvid: @raulk thoughts?

mvid: @raulk wrote it back to use sync.Mutex, I don't want this to get stuck

README.md
```diff
@@ -8,7 +8,7 @@
 [![Build Status](https://travis-ci.org/libp2p/go-libp2p-core.svg?branch=master)](https://travis-ci.org/libp2p/go-libp2p-core)
 [![Discourse posts](https://img.shields.io/discourse/https/discuss.libp2p.io/posts.svg)](https://discuss.libp2p.io)
 
-> Home to the interfaces and abstractions that make up go-libp2p.
+> Home to the interfaces and abstractions that make up go-libp2p. Action taking place in the [genesis PR](https://github.com/libp2p/go-libp2p-core/pull/1).
```
raulk: You're re-adding a line from history.

go.mod
```diff
@@ -13,5 +13,6 @@ require (
 	github.com/multiformats/go-multiaddr v0.0.2
 	github.com/multiformats/go-multihash v0.0.1
 	github.com/spacemonkeygo/openssl v0.0.0-20181017203307-c2dcc5cca94a
-	golang.org/x/crypto v0.0.0-20190225124518-7f87c0fbb88b
+	go.opencensus.io v0.21.0
+	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
```
raulk: Let's not slip in unrelated dependency upgrades.

metrics/register.go
```diff
@@ -0,0 +1,71 @@
...
+// LookupViews returns all views for a Namespace name. Returns an error if the
+// Namespace has not been registered.
+func LookupViews(name string) ([]*view.View, error) {
+	mutex.Lock()
+	views, ok := registeredViews[name]
+	mutex.Unlock()
+	if !ok {
+		return nil, ErrUnregisteredNamespace{Namespace: name}
+	}
+	return views, nil
```
raulk: You want to copy this slice to a new one to protect your state.

metrics/register.go
```diff
@@ -0,0 +1,71 @@
...
+// with opencensus and maintained in the global registered views map
+func RegisterViews(namespace string, views ...*view.View) error {
+	mutex.Lock()
+	_, ok := registeredViews[namespace]
+	if !ok {
+		return ErrDuplicateNamespaceRegistration{Namespace: namespace}
+	}
+
+	if err := view.Register(views...); err != nil {
+		delete(registeredViews, namespace)
```
raulk: Are we even adding anything to `registeredViews`?

raulk:
@mvid – could you add unit tests for this, please?

go.mod
mvid: This is a result of running `go mod tidy`, at request of @bigs. Cole, I'm undoing that

metrics/register.go
mvid: Sorry, this was an oversight from switching back from sync.Map. Corrected

metrics/register.go
mvid: Done

go.mod
bigs: @raulk yeah, not super sure how to handle things like this. in my experience doing any work with the `go` tool is liable to add things to `go.mod` as is using the language server, etc. i've had him `go mod tidy` before commits, but these things seem to bleed in. i think they're okay, as the mental overhead of pruning them would seem to be a lot.

README.md
mvid: Strange. Removed

metrics/register_test.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"fmt"
+	"go.opencensus.io/stats"
+	"testing"
+
+	"go.opencensus.io/stats/view"
+)
+
+func testWrapper(t *testing.T, f func(t *testing.T)) func(t *testing.T) {
+	for k, _ := range registeredViews {
```
bigs:
```golang
registeredViews = make(map[string][]*view.View)
```

should do it

metrics/register_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func newTestMeasure(name string) stats.Measure {
+	return stats.Int64(fmt.Sprintf("test/measure/%s", name),
+		fmt.Sprintf("Test measurement %s", name),
+		stats.UnitDimensionless,
+	)
+}
+
+func newTestView(name string) view.View {
```
bigs: this should just return `*view.View` i think

bigs:
cleaned up in #19

PR #15, opened 2019-05-31
feat(peer): implement AddrInfosFromP2pAddrs and SplitAddr

Stebalien:
* SplitAddr is a simpler way to split an address into a multiaddr and an ID.
* AddrInfosFromP2pAddrs converts a set of multiaddrs into a set of AddrInfos.


Stebalien:
(fixed test coverage and a bug)

PR #17, opened 2019-06-19
eventbus abstraction + initial events

whyrusleeping:
Still waiting on https://github.com/libp2p/go-eventbus/pull/1 to merge, but getting everything else ready in the meantime.


event/doc.go
```diff
@@ -0,0 +1,11 @@
+// Package event contains the abstractions for a local event bus, along with the standard events
+// that libp2p subsystems may emit.
+//
+// Source code is arranged as follows:
+// 	* doc.go: this file.
```
Kubuxu:
```suggestion
// * doc.go: this file.
```

raulk: I want this to be a code block, otherwise the "list" formatting will get messed up.

PR #18, opened 2019-06-19
remove dependency on go-eventbus.

raulk:


whyrusleeping:
Making emitter a type like that means that eventbus has to depend on this package. Im not really against that at this point, just want to point it out

raulk:
@whyrusleeping yeah, I was struggling even with `EmitFunc` being a function type, because the type had a `Close()` method that prevented extracting an abstraction. We've succumbed to making go-eventbus depend on go-libp2p-core for now.

PR #19, opened 2019-06-19
Add OpenCensus metrics registration functionality to core

bigs:
Commit cleanup of #4 which got mangled when consolidate-skeleton was merged.


lanzafame:
@bigs as we discussed in Barcelona, this really isn’t how registration of OC metrics should be done. Can we look at reverting this?

raulk:
Hey @lanzafame! Would you mind posting the gist of your offline convo? I’d like to know what the recommendation is so I can weigh in.

lanzafame:
Registration is designed to be done by the end user and this is just putting utilities into the core interface package for no real benefit.

PR #20, opened 2019-06-19
peerstore/protobook: support removing protocols.

raulk:
Necessary to apply protocol deltas from identify push (#17).


raulk:
Gonna merge this as it's trivial and I'm the owner of the peerstore anyway.

PR #22, opened 2019-06-21
eventbus interface changes

magik6k:
See https://github.com/libp2p/go-eventbus/pull/9

### DIFF
--- a/event/bus.go
+++ b/event/bus.go
@@ -22,39 +22,44 @@ type Emitter interface {
 	Emit(evt interface{})
 }
 
+// Subscription represents a subscription to one or multiple event types.
 type Subscription interface {
+	io.Closer
+
+	// Out returns the channel from which to consume events.
 	Out() <-chan interface{}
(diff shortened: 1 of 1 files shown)


event/bus.go
```diff
@@ -30,11 +35,14 @@ type Bus interface {
...
-	// defer cancel()
-	Subscribe(typedChan interface{}, opts ...SubscriptionOpt) (CancelFunc, error)
+	// sub, err := eventbus.Subscribe(new(EventType))
+	// defer sub.Close()
+	// for e := range sub.Out() {
+	//   event := e.(EventType) // guaranteed safe
+	//   [...]
+	// }
+	// TODO: update doc
+	Subscribe(eventType interface{}, opts ...SubscriptionOpt) (Subscription, error)
```
Kubuxu: Shouldn't this be `[]interface{}`.

Ahh, it accepts both.
This should be documented.

PR #25, opened 2019-06-25
event: Add autonat events

magik6k:
Dependency for generating autonat events


event/autonat.go
```diff
@@ -0,0 +1,5 @@
+package event
```
raulk: Could we change this filename to something more abstract that doesn't reference a particular subsystem? `routability.go`?

PR #27, opened 2019-06-28
avoid marshaling the protobuf on compare

Stebalien:
We also should consider keeping these keys in their serialized form.


Stebalien:
Replaced by #28.

PR #28, opened 2019-06-28
fix: do not allocate when comparing keys

dignifiedquire:


crypto/rsa_go.go
```diff
@@ -93,7 +108,35 @@ func (sk *RsaPrivateKey) Raw() ([]byte, error) {
...
+	b := other.sk
+
+	if a.PublicKey.N.Cmp(b.PublicKey.N) != 0 {
+		return false
+	}
+	if a.PublicKey.E != b.PublicKey.E {
+		return false
+	}
+
+	if a.D.Cmp(b.D) != 0 {
```
dignifiedquire: @Stebalien not sure if we should do this, I would argue if the public key matches, we can assume the private key matches

dignifiedquire:
> Introduces a recursive call to KeyEquals in the openssl implementation.

where is that impl?

dignifiedquire:
> Introduces a recursive call to KeyEquals in the openssl implementation.

@Stebalien should we just delete `KeyEquals`, and embed the logic into the impls?

crypto/rsa_go.go
Stebalien: I'd just do it. It shouldn't be expensive.

dignifiedquire:
nvm, fixed openssl, it has to allocate unfortuantely as the lib does not expose a way to compare keys :(

Stebalien:
> @Stebalien should we just delete KeyEquals, and embed the logic into the impls?

Let's deprecate it for now (we use it).

crypto/key.go
```diff
@@ -345,7 +344,9 @@ func KeyEqual(k1, k2 Key) bool {
 		return true
 	}
 
-	b1, err1 := k1.Bytes()
-	b2, err2 := k2.Bytes()
-	return bytes.Equal(b1, b2) && err1 == err2
+	if k1.Type() != k2.Type() {
+		return false
+	}
```
Stebalien: Let's move the type check into the `Equals` implementation. I believe we call that manually in some places.

crypto/openssl_common.go
```diff
@@ -94,5 +104,13 @@ func (sk *opensslPrivateKey) Raw() ([]byte, error) {
...
 func (sk *opensslPrivateKey) Equals(k Key) bool {
-	return KeyEqual(sk, k)
+	a, err := sk.Raw()
+	if err != nil {
+		return false
+	}
+	b, err := k.Raw()
+	if err != nil {
+		return false
+	}
```
Stebalien: helper?

dignifiedquire:
benches copied from #29 

```
# this branch
BenchmarkRSAKeyEqualsCheck-12    	100000000	        15.7 ns/op
# master
BenchmarkRSAKeyEqualsCheck-12    	  100000	     11345 ns/op
```

dignifiedquire:
openssl

```
BenchmarkRSAKeyEqualsCheck-12    	10000000	       133 ns/op
```

crypto/openssl_common.go
dignifiedquire: meh

Stebalien: Well, I fixed it anyways... (mostly so I could consistently fix https://github.com/libp2p/go-libp2p-core/pull/28#discussion_r298548461)

dignifiedquire: thanks

crypto/rsa_go.go
dignifiedquire: I think we should remove it, openssl only checks the public key, and this check is not constant time, so it could leak information about the private key afict

Stebalien: Meh, fine by me either way.

Stebalien: Done!

PR #29, opened 2019-06-28
Feat/rsa caching

whyrusleeping:
This is making GC realllly hot on the machines i've been watching.

Here's the before and after benchmark results:

Before:
```
why@WhyNet ~/g/s/g/l/g/crypto> go test -run XXXX -bench=BenchmarkRSAKey
goos: darwin
goarch: amd64
pkg: github.com/libp2p/go-libp2p-core/crypto
BenchmarkRSAKeyEqualsCheck-4   	  100000	     13415 ns/op	    5443 B/op	     118 allocs/op
PASS
ok  	github.com/libp2p/go-libp2p-core/crypto	4.706s

why@WhyNet ~/g/s/g/l/g/crypto> go test -run XXXX -bench=BenchmarkRSAKey -tags=openssl
# github.com/libp2p/go-libp2p-core/crypto.test
ld: warning: directory not found for option '-L/usr/local/opt/openssl@1.1/lib'
goos: darwin
goarch: amd64
pkg: github.com/libp2p/go-libp2p-core/crypto
BenchmarkRSAKeyEqualsCheck-4   	  100000	     14560 ns/op	    5024 B/op	      18 allocs/op
PASS
ok  	github.com/libp2p/go-libp2p-core/crypto	2.789s
```

After:

```
why@WhyNet ~/g/s/g/l/g/crypto> go test -run XXXX -bench=BenchmarkRSAKey
goos: darwin
goarch: amd64
pkg: github.com/libp2p/go-libp2p-core/crypto
BenchmarkRSAKeyEqualsCheck-4   	20000000	        69.6 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/libp2p/go-libp2p-core/crypto	6.666s

why@WhyNet ~/g/s/g/l/g/crypto> go test -run XXXX -bench=BenchmarkRSAKey -tags=openssl
# github.com/libp2p/go-libp2p-core/crypto.test
ld: warning: directory not found for option '-L/usr/local/opt/openssl@1.1/lib'
goos: darwin
goarch: amd64
pkg: github.com/libp2p/go-libp2p-core/crypto
BenchmarkRSAKeyEqualsCheck-4   	30000000	        52.5 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/libp2p/go-libp2p-core/crypto	3.788s
```

### DIFF
--- a/crypto/key.go
+++ b/crypto/key.go
@@ -282,7 +282,20 @@ func UnmarshalPublicKey(data []byte) (PubKey, error) {
 		return nil, ErrBadKeyType
 	}
 
-	return um(pmes.GetData())
+	pk, err := um(pmes.GetData())
+	if err != nil {
+		return nil, err
+	}
+
+	switch tpk := pk.(type) {
+	case *RsaPublicKey:
+		bcopy := make([]byte, len(data))
+		copy(bcopy, data)
+
+		tpk.cached = bcopy
+	}
+
+	return pk, nil
 }
 
 // MarshalPublicKey converts a public key object into a protobuf serialized
--- a/crypto/rsa_go.go
+++ b/crypto/rsa_go.go
@@ -9,6 +9,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"io"
+	"sync"
 
 	pb "github.com/libp2p/go-libp2p-core/crypto/pb"
 
@@ -23,6 +24,9 @@ type RsaPrivateKey struct {
 // RsaPublicKey is an rsa public key
 type RsaPublicKey struct {
 	k rsa.PublicKey
+
+	cacheLk sync.Mutex
+	cached  []byte
 }
 
 // GenerateRSAKeyPair generates a new rsa private and public key
@@ -35,7 +39,7 @@ func GenerateRSAKeyPair(bits int, src io.Reader) (PrivKey, PubKey, error) {
 		return nil, nil, err
 	}
 	pk := priv.PublicKey
-	return &RsaPrivateKey{sk: *priv}, &RsaPublicKey{pk}, nil
+	return &RsaPrivateKey{sk: *priv}, &RsaPublicKey{k: pk}, nil
 }
 
 // Verify compares a signature against input data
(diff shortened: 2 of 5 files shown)


crypto/key.go
```diff
@@ -282,7 +282,20 @@ func UnmarshalPublicKey(data []byte) (PubKey, error) {
 		return nil, ErrBadKeyType
 	}
 
-	return um(pmes.GetData())
+	pk, err := um(pmes.GetData())
+	if err != nil {
+		return nil, err
+	}
+
+	switch tpk := pk.(type) {
```
Stebalien: :scream:

crypto/rsa_go.go
```diff
@@ -23,6 +24,9 @@ type RsaPrivateKey struct {
 // RsaPublicKey is an rsa public key
 type RsaPublicKey struct {
 	k rsa.PublicKey
+
+	cacheLk sync.Mutex
```
Stebalien: What if we just _always_ cache? No lock.

crypto/key.go
Stebalien: Do we need to cache the entire protobuf or can we just cache (and compare) the raw data.

whyrusleeping: if we do that, then we end up still having to create a whole protobuf every time we call bytes. The problem here is with our interfaces, we have asymetry between whats given to the key to unmarshal itself, and what we expect it to give us when we ask it for its bytes.

crypto/rsa_go.go
whyrusleeping: what do you mean?

crypto/key.go
Stebalien: > if we do that, then we end up still having to create a whole protobuf every time we call bytes.

Is that a common operation (except when comparing)?

crypto/rsa_go.go
Stebalien: Always compute the serialized public key and store it so `Bytes()` becomes a no-op `key.cache`.

whyrusleeping: the problem i have there is that we get a public key from the private key, and that public key doesnt get cached in the private key (we could, but thats proliferating a bit). So then every time you would call .PublicKey on a PrivateKey, you'd end up paying for that serialization. maybe not the worst thing though...

Stebalien:
https://github.com/libp2p/go-libp2p/pull/665

PR #31, opened 2019-07-10
Remove support for blowfish

bigs:
Removes a reference to blowfish

### DIFF
--- a/crypto/key.go
+++ b/crypto/key.go
@@ -191,6 +191,8 @@ func KeyStretcher(cipherType string, hashType string, secret []byte) (StretchedK
 	case "AES-256":
 		ivSize = 16
 		cipherKeySize = 32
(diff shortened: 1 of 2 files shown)


crypto/key.go
```diff
@@ -191,10 +191,6 @@ func KeyStretcher(cipherType string, hashType string, secret []byte) (StretchedK
 	case "AES-256":
 		ivSize = 16
 		cipherKeySize = 32
-	case "Blowfish":
```
Stebalien: We should have a default case, right?

Kubuxu: We for sure need a default case here.

bigs: aye, can do

bigs:
bump @kubuxu

PR #32, opened 2019-07-10
Replace bytes.Equal -> subtle.ConstantTimeCompare

bigs:
Removes reliance on short-circuiting equality comparisons which can be cryptographically unsound.

### DIFF
--- a/crypto/ed25519.go
+++ b/crypto/ed25519.go
@@ -1,7 +1,7 @@
 package crypto
 
 import (
-	"bytes"
+	"crypto/subtle"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


crypto/ed25519.go
```diff
@@ -131,7 +131,7 @@ func UnmarshalEd25519PrivateKey(data []byte) (PrivKey, error) {
 		// Remove the redundant public key. See issue #36.
 		redundantPk := data[ed25519.PrivateKeySize:]
 		pk := data[ed25519.PrivateKeySize-ed25519.PublicKeySize : ed25519.PrivateKeySize]
-		if !bytes.Equal(pk, redundantPk) {
+		if subtle.ConstantTimeCompare(pk, redundantPk) != 1 {
```
Stebalien: This is a _public_ key comparison so we don't need to do this.

crypto/ed25519.go
```diff
@@ -105,7 +105,7 @@ func (k *Ed25519PublicKey) Equals(o Key) bool {
 		return false
 	}
 
-	return bytes.Equal(k.k, edk.k)
+	return subtle.ConstantTimeCompare(k.k, edk.k) == 1
```
Stebalien: Also a public key comparison.

bigs:
reverted to faster comparisons for pubkeys. going to merge once tests pass.

PR #33, opened 2019-07-10
Return error rather than panic in GenerateEKeyPair

bigs:
Stops us from panicking in the event that we receive a curve we don't support.


Stebalien:
Can we add a test for this?

bigs:
tested @Stebalien

PR #37, opened 2019-07-11
update insecure transport to plaintext/2.0.0

### LINKED ISSUE #36: adjust `insecure` secure channel to adhere to new spec
We have now [merged the spec](https://github.com/libp2p/specs/pull/186) for the `insecure` channel and we need to implement it to (a) fix downstream bugs, (b) enable interoperability testing with py-libp2p and jvm-libp2p.

This is where it's at: https://github.com/libp2p/go-libp2p-core/blob/master/sec/insecure/insecure.go

CC @mhchia @Nashatyrev.

yusefnapora:
This changes the behavior of the insecure "security" transport to conform to the [new spec](https://github.com/libp2p/specs/pull/186). Will close #36.

@bigs I can't seem to figure out how to correctly import the `crypto.pb` file from the new `plaintext.pb` file I added.

The way it's set up now, `protoc` generates an import statement like

```go
import (
  "other/stuff"
  pb "crypto/pb"
)
```

Which doesn't resolve to anything. I manually changed that to `pb "github.com/libp2p/go-libp2p-core/crypto/pb"`, which is what's committed in here, just so I could get something building.

I also copied the test code over from secio and stripped out the stuff related to actual crypto. It seems like it works...

Another thing I'm not sure about is whether adding a new private key argument to the `New` constructor will break anything. I think I remember that the reflection magic thing will inject a private key if you take one in as an arg, but I can't remember where that code is, lol.


yusefnapora:
I'm also wondering if we should log a warning when the transport is instantiated, just as a guard against people accidentally enabling this during production. I don't know anything about go logging though, so idk what's appropriate for a library to log and so on.

yusefnapora:
About the protobuf import thing, I've discovered that it basically works if you add the `go_package` option to your protobuf files like so:

crypto.proto:

```protobuf
syntax = "proto2";

package crypto.pb;
option go_package = "github.com/libp2p/go-libp2p-core/crypto/pb";
```

Then when you import `crypto.proto` from `plaintext.proto`, the generated code in `plaintext.pb.go` is:

```go
import (
	fmt "fmt"
	proto "github.com/gogo/protobuf/proto"
	pb "github.com/libp2p/go-libp2p-core/crypto/pb"
	io "io"
	math "math"
)
```

Which is great and the compiler loves me. The only problem is that, instead of generating `crypto.pb.go` in the current directory, `protoc` now makes a directory tree, e.g. 

```
.
└── github.com
    └── libp2p
        └── go-libp2p-core
            └── sec
                └── insecure
                    └── pb
                        └── plaintext.pb.go
```

So that's annoying. It's not hard to work around - we can `find` the generated file in the tree and move back where it's supposed to go.

Maybe there's a way that's less of a hack though, i don't know.

sec/insecure/insecure.go
```diff
@@ -36,33 +42,163 @@ func (t *Transport) LocalPeer() peer.ID {
...
+	}
+
+	// Add remote ID and key to conn state
+	ic.remotePubKey = remotePubkey
+	ic.remote = remoteID
+	return nil
+}
+
+// read and write a message at the same time.
+func readWriteMsg(c msgio.ReadWriter, out []byte) ([]byte, error) {
```
bigs: even though it makes the function quite long, this can be inlined

sec/insecure/insecure.go
```diff
@@ -36,33 +42,163 @@ func (t *Transport) LocalPeer() peer.ID {
...
+	wresult := make(chan error)
+	go func() {
```
bigs: this coroutine should be unnecessary

sec/insecure/insecure.go
```diff
@@ -36,33 +42,163 @@ func (t *Transport) LocalPeer() peer.ID {
...
+	}
+
+	return &pb.Exchange{
+		Id:     []byte(id),
+		Pubkey: pubkey,
+	}, nil
+}
+
+func (ic *Conn) runHandshakeSync(ctx context.Context) error {
+	insecureM := msgio.NewReadWriter(ic.Conn)
```
bigs: not entirely necessary but could use the `ggio` delimited reader and writer pair, which will accept any `proto.Message`. would save you from having to marshal/unmarshal

sec/insecure/insecure.go
```diff
@@ -36,33 +42,163 @@ func (t *Transport) LocalPeer() peer.ID {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if p != conn.remote {
+		return nil, fmt.Errorf("remote peer sent unexpected peer ID. expected=%s received=%s",
+			p, conn.remote)
+	}
+
+	if !p.MatchesPublicKey(conn.remotePubKey) {
```
bigs: this should be redundant given the checks below

yusefnapora:
Sweet, thanks for the feedback @bigs, it's much better now :)

sec/insecure/insecure.go
```diff
@@ -5,64 +5,172 @@ package insecure
...
+	}
+
+	return &pb.Exchange{
+		Id:     []byte(id),
+		Pubkey: pubkey,
+	}, nil
+}
+
+func (ic *Conn) runHandshakeSync(ctx context.Context) error {
+	const maxSize = 1 << 16
```
bigs: we have a default here https://github.com/libp2p/go-libp2p-core/blob/master/network/network.go#L23

but honestly we can probably place a smaller limit on key exchanges. just linking for info

bigs: on second thought, may be worth just using that constant. also this is reminding me, one of msgio's big benefits is buffer pooling, but since this is a test-only upgrader, we're probably fine with a little extra allocation.

yusefnapora: awesome, thanks for pointing me at the constant. it seems best to use it for consistency rather than have more magic numbers sprinkled about.

sec/insecure/insecure.go
```diff
@@ -5,64 +5,172 @@ package insecure
...
+// No authentication of the remote identity is performed.
 type Transport struct {
-	id peer.ID
+	id  peer.ID
+	key ci.PrivKey
 }
 
 // New constructs a new insecure transport.
-func New(id peer.ID) *Transport {
+func New(id peer.ID, key ci.PrivKey) *Transport {
```
raulk: **This causes pretty severe API breakage.**

1. in the `Transport` option of the go-libp2p constructor.
2. elsewhere (e.g. go-tcp-transport) where this constructor is used directly.

raulk: The private key is unnecessary here. The public key should do.

raulk: We should deprecate the `New` function and make a `WithIdentity()` function that the public key, and returns a function of signature `func(id peer.ID) *Transport` that can then be used with the go-libp2p constructor.

bigs:
definite lapse here, sorry. i think we should just add a new version of the protocol alongside 1.0.0. that’s what the versioning is for and this protocol is different.

raulk:
@bigs I thought of that too, but I concluded that v1 does not really work, and there’s no point ossifying something that’s broken.

sec/insecure/insecure.go
yusefnapora: Adding a new constructor function and deprecating the old is definitely the way to go 👍

I added the private key to avoid returning `nil` from `LocalPrivateKey` - if we don't care about that then we can just take the public key.

PR #40, opened 2019-07-15
Add API compatibity check to travis.

### LINKED ISSUE #39: Investigate using tooling to prevent compatibility breakages
There exist tooling that will detect compatibility breaking changes. It would alert us in cases like recently when a new version of insecure transport or addition of event bus broke the compatibility.
One of the tools is https://github.com/smola/gocompat
The idea would be to run it in CI for every PR and compare the compatibility with the latest release/master.

Regarding compatibility: https://blog.merovius.de/2015/07/29/backwards-compatibility-in-go.html is worth a read.

Would help in cases like #23 #37 #38 https://github.com/libp2p/go-libp2p/issues/678

Kubuxu:
Resolves #39


yusefnapora:
As the author of #37, this seems awesome to me 😄 There are plenty of reasons to break API compatibility, but it's always nice to at least know that you're doing it

tools.go
```diff
@@ -0,0 +1,7 @@
+// +build tools
+
+package core
+
+import (
+	_ "github.com/smola/gocompat"
```
raulk: I read the usage instructions on the `gocompat` repo, but I'm not sure I follow why we need an import for side-effects here. Isn't the CLI tool self-contained?

Stebalien: It's how one tricks `go mod` into tracking it. That way, one can run `go run github.com/smola/gocompat/cmd/gocompat` and always get the right version.

Kubuxu: The build that is made such that it isn't used for the build.

raulk: What a hack... *sigh* `scope=test` in Maven, `devDependencies` in npm.

.travis.yml
```diff
@@ -17,8 +17,15 @@ env:
...
 
-script:
-  - bash <(curl -s https://raw.githubusercontent.com/ipfs/ci-helpers/master/travis-ci/run-standard-tests.sh)
+script: bash <(curl -s https://raw.githubusercontent.com/ipfs/ci-helpers/master/travis-ci/run-standard-tests.sh)
+
+jobs:
+  include:
+    - stage: "Test"
+      script: "./compat-check"
+      name: "Compatibility Test"
```
raulk:
```suggestion
      name: "API compatibility test"
```

Kubuxu:
I've removed the breaking change and changed the line.

Kubuxu:
Closes https://github.com/libp2p/go-libp2p-core/issues/39

PR #41, opened 2019-07-15
feat: support encoding/decoding peer IDs as CIDs in text

Stebalien:
What:

1. Supports decoding CIDs (of libp2p keys) as peer IDs.
2. Continues to encode peer IDs as base58 multihashes by default.
3. Adds functions for converting between peer IDs and CIDs.
4. Deprecates IDB58{Decode,Encode}, replacing them with {Decode,Encode}.

Why:

1. We _need_ to support multibase somehow so we can put peer IDs in domains.
2. This makes peer IDs fully self describing. That is, the CID itself indicates that it's a hash of a libp2p public key.
3. It's much easier to upgrade wire protocols than text. This change punts
pids-as-cids on the wire down the road but that's something we can revisit if it ever becomes relevant. 

See https://github.com/libp2p/specs/issues/111 for context.

Deviations from that issue:

* This _retains_ the current peer ID inlining of ed25519 keys. That turned out to be a nightmare, one I'd like to punt a bit more down the road.
* Likewise, this _punts_ the question of embedding the multi-hash algorithm in the public key.

---

Alternative: Hack this into go-ipfs and make it _only_ valid in `/ipns` links. Honestly, we can do that first and then merge this later.


hsanjuan:
> Continues to encode peer IDs as base58 multihashes by default.

Why don't we change this default now already?

jacobheun:
This change makes sense to me. If we change the default now we wouldn't give the network any time to upgrade. This would cause the node to announce its peer id as a cid hash wouldn't it?

One question, right now we're using the `libp2p-key` multicodec. Would we get any benefit for being more granular with this to know what type of libp2p key it is?

Stebalien:
@hsanjuan 
> Why don't we change this default now already?

Compatibility.

---

@jacobheun 

>  This would cause the node to announce its peer id as a cid hash wouldn't it?

We're not changing the wire format, just the format in text.

> Would we get any benefit for being more granular with this to know what type of libp2p key it is?

The key already specifies that internally. libp2p-key just means that we're pointing to a protobuf with the structure `{type: int, data: [bytes]}`.

jacobheun:
Great, thanks for clarifying. 👍

hsanjuan:
> Compatibility.

OK, but is there an upgrade timeline? It says `in the future` which is not super helpful... also there does not seem to be a way to enforce CID-encoding (other than going manual with `ToCid().String()` so it is hard to test if anything breaks.

If we are suppose to move from IDB58Encode to Encode all over our code, and libp2p will one day flip the switch for us, then it may just flip the switch now in the Encode function. Why not making the move from `IDB58Encode` to `Encode` should be actual `opt-in` to work with CIDs?

Stebalien:
This is a two step upgrade:

1. Add support for decoding CIDs as peer IDs everywhere.
2. Flip the switch and start encoding them that way. When we do that, I'd like to make IDB58Encode _also_ encode peer IDs as CIDs. The only reason I'm deprecating it in favor of Encode is that the function name IDB58Encode will be incorrect.

If we start encoding as CIDs _now_, peers using the old version won't be able to understand peer IDs encoded by the new version. This won't break the network but it _will_ confuse users.

> OK, but is there an upgrade timeline?

Sure. How about 3 months after the last libp2p implementation upgrades and releases?

lidel:
PSA: proposed change to libp2p specs to default text representation to CID in Base32:
- https://github.com/libp2p/specs/pull/209 (RFC 0001: Text Peer Ids as CIDs)

peer/peer.go
```diff
@@ -155,10 +158,66 @@ func IDHexDecode(s string) (ID, error) {
 }
 
 // IDHexEncode returns the hex-encoded multihash representation of the ID.
+//
+// Deprecated: Don't raw-hex encode peer IDs, use base16 CIDs.
 func IDHexEncode(id ID) string {
```
Stebalien: We should _really_ remove these. But that's a breaking change and we can do that later.

peer/peer.go
```diff
@@ -155,10 +158,66 @@ func IDHexDecode(s string) (ID, error) {
...
+		}
+		return "", fmt.Errorf("can't convert CID of type %s to a peer ID", s)
+	}
+	return ID(c.Hash()), nil
+}
+
+// ToCid encodes a peer ID as a CID of the public key.
+//
+// If the peer ID is invalid (e.g., empty), this will return the empty CID.
+func ToCid(id ID) cid.Cid {
```
raulk: nit: go style is to capitalise acronyms. We should create an alias `cid.CID => cid.Cid`.

```suggestion
func ToCID(id ID) cid.Cid {
```

peer/peer.go
```diff
@@ -155,10 +158,66 @@ func IDHexDecode(s string) (ID, error) {
...
+// Encode encodes a peer ID as a string.
+//
+// At the moment, it base58 encodes the peer ID but, in the future, it will
+// switch to encoding it as a CID by default.
+func Encode(id ID) string {
+	return IDB58Encode(id)
+}
+
+// FromCid converts a CID to a peer ID, if possible.
+func FromCid(c cid.Cid) (ID, error) {
```
raulk:
```suggestion
func FromCID(c cid.Cid) (ID, error) {
```

Stebalien:
I'm fine with those changes but they _are_ inconsistent with the rest of go-ipfs/go-libp2p which we're not going to change as following a style guideline isn't worth the confusion of having two ways to do the same thing.

peer/peer.go
lidel: @raulk  I don't think we should introduce new convention here.
All our codebases already use things like [`NewCidV1`](https://github.com/search?q=org%3Aipfs+NewCidV1&type=Code) and [`FromCid`](https://github.com/search?q=org%3Aipfs+FromCID&type=Code). 

If we choose to change them to use capitalization, that needs to happen across all repos (in separate, coordinated PRs), otherwise we just  introduce noise by mixing conventions.

lidel:
@raulk Would it be ok to merge as-is? :pray: 
I'd like to keep PRs with CID changes small and on topic. See my note in https://github.com/libp2p/go-libp2p-core/pull/41#discussion_r355518175

raulk:
Go ahead!

lidel:
@Stebalien @raulk I am missing permission to push to this branch – mind adding me, or applying below fix youself?

```
ERROR: Permission to libp2p/go-libp2p-core.git denied to lidel.
fatal: Could not read from remote repository.
```

The only conflict with `master` is go version in `.travis.yml`, the fix is:

```diff
-   - 1.12.x
+   - 1.13.x
```

Stebalien:
Welcome to the go-libp2p team!

PR #42, opened 2019-07-16
bring back plaintext 2.0.0 with new constructor

yusefnapora:
This is #37 again, but instead of adding an extra argument to the `insecure.New` constructor function I just added a new `insecure.NewWithIdentity` function that accepts the extra key argument.

I _think_ the `Transport` option in go-libp2p will inject the key for me, according to [the comments](https://github.com/libp2p/go-libp2p/blob/master/options.go#L55). Places that are calling the constructor directly (like the go-tcp-transport tests) can update after this merges.

This change is a bit awkward, since calling the old constructor will partially initialize the `Transport` struct and leave it without a `key`. If that happens, we just skip the key exchange entirely and behave like plaintext/1.0.0, but this will cause the connection to hang until timeout if one side was initialized with a key (and is expecting to receive one over the wire) and one side does not have a key. It should work correctly (like plaintext 1.0) if both sides use the old constructor, so the tcp tests should still pass.

@raulk on the earlier PR you mentioned we should only accept the public key - that's all we need to send on the wire, but the `SecureTransport` interface has a `LocalPrivateKey` method. I wasn't sure how to know whether anything depends on getting a non-nil value. I can switch to just storing the local public key; I just wanted to run that by you first.

I'd like to test this with the workspace, but I'm getting what seem to be unrelated failures building at the moment...

When I do a new recursive clone of the workspace and do `./workspace.sh local`, then run `git submodule foreach 'go build ./...'` I get a failure:

```
git submodule foreach 'go build ./...'
Entering 'go-addr-util'
Entering 'go-buffer-pool'
Entering 'go-conn-security-multistream'
Entering 'go-eventbus'
Entering 'go-flow-metrics'
Entering 'go-libp2p'
Entering 'go-libp2p-autonat'
Entering 'go-libp2p-autonat-svc'
Entering 'go-libp2p-blankhost'
Entering 'go-libp2p-circuit'
Entering 'go-libp2p-connmgr'
Entering 'go-libp2p-consensus'
Entering 'go-libp2p-core'
Entering 'go-libp2p-daemon'
go: downloading github.com/prometheus/client_golang v0.9.3
go: downloading github.com/ipfs/go-datastore v0.0.5
go: downloading golang.org/x/xerrors v0.0.0-20190513163551-3ee3066db522
go: downloading github.com/lucas-clemente/quic-go v0.11.2
go: downloading github.com/libp2p/go-libp2p-routing v0.1.0
build github.com/libp2p/go-libp2p-daemon/p2p-keygen: cannot load github.com/ipfs/go-datastore: reading github.com/ipfs/go-datastore/go.mod at revision v0.0.5: git -c protocol.version=0 fetch --unshallow -f https://github.com/ipfs/go-datastore refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /Users/yusef/go/pkg/mod/cache/vcs/516390995ce8062700fa248e29178f57c9d99d40ed4998a68a87ee8153f70574: exit status 128:
	fatal: not a git repository: '.git'
fatal: run_command returned non-zero status for go-libp2p-daemon
.
```

I'll poke at it some more. 

@bigs, any ideas?


bigs:
@yusefnapora this should be okay. that's a helper binary, and it compiles fine on master. perhaps try compiling it manually, i.e.

```bash
cd path/to/p2p-keygen
go build
```

and also try updating it to the latest released tag, v0.2.0, and building there. let me know!

sec/insecure/insecure.go
```diff
@@ -5,64 +5,188 @@ package insecure
...
+// peer presents as their ID and public key.
+// No authentication of the remote identity is performed.
 type Transport struct {
-	id peer.ID
+	id  peer.ID
+	key ci.PrivKey
 }
 
 // New constructs a new insecure transport.
+// This function is deprecated. Please use NewWithIdentity instead.
```
Kubuxu: Use `Deprecated:` prefix in the comment so it is detected by tooling.

raulk:
@yusefnapora – any chance you can tackle follow-ups today?

yusefnapora:
@raulk sure, thanks for the bump 👍

yusefnapora:
I figured out why the tests for go-tcp-transport were failing (when calling the old constructor) & fixed in https://github.com/libp2p/go-libp2p-core/pull/42/commits/23f3b50406a99caf3ed602571c027bef6b6194e0 - I think this is safe to merge now.

I started making branches for go-libp2p and go-tcp-transport to use the new constructors, but I got derailed by the tcp test failures and parked them. I'll push those up in a little bit once I make sure they're all good.

yusefnapora:
Hmm... maybe not quite ready yet 😄 - I'm now seeing weird behavior when the TCP tests do use the new constructor, so I'm going to investigate a bit more and make sure the handshake actually works in practice.

yusefnapora:
I think that the handshake was deadlocking; switching back to the simultaneous read / write method of exchanging messages that I stole from secio seems to have fixed it.

bigs:
ah that was my bad! if the test is happening in a single thread it’s def a
problem.

yusefnapora:
okay, finally gonna merge, since I've got a good feeling about this one :)

I've got branches for `go-libp2p`, `go-tcp-transport`, `go-ws-tranport` and `go-libp2p-transport-upgrader` that use the new constructor, and I verified that the tests for each of those pass with this implementation whether you use the old or new constructors. I'll push those branches up in a few mins on their respective repos.

PR #44, opened 2019-07-22
avoid duplicate randomly generated keys/peer-ids

### LINKED ISSUE #43: Duplicate Randomly Generated PeerIDs
https://github.com/libp2p/go-libp2p-core/blob/79bf30a5a421b6f0d304c411b78c391fd9f1ca46/test/peer.go#L14-L30

`RandPeerID` (and therefore `RandPeerIDFatal` functions can return the same peerID on systems that do not fully support `time`'s `UnixNano()` (e.g. `time.Now().UnixNano()`). See golang/go#27301 for details.

There are any number of ways to solve this including:
1) Replace `time.Now().UnixNano()` with `math/rand`'s `rand.Int63()`
2) Use a workaround increment counter as in: https://github.com/libp2p/go-libp2p-core/blob/79bf30a5a421b6f0d304c411b78c391fd9f1ca46/test/crypto.go#L11-L20
3) Use a global locked random reader as `math/rand` does internally https://github.com/golang/go/blob/f518a96e009157c51f8ecf619032bbb7873228f9/src/math/rand/rand.go#L288
4) Stop using non-seeded randomness in our tests so that they're actually reproducible (e.g. use random seeds controlled by the tests instead of by time)

My personal favorites are 1 and 4. However, whatever gets this issue fixed ASAP works for me. I'm also happy to implement the changes as long as I get pre-approval on which method (1,2,3,4, or your favorite method here) to use.

@raulk @Stebalien

Stebalien:
This implements #4 from #43.

fixes #43

### DIFF
--- a/test/crypto.go
+++ b/test/crypto.go
@@ -3,20 +3,16 @@ package test
 import (
 	"math/rand"
 	"sync/atomic"
-	"time"
 
 	ci "github.com/libp2p/go-libp2p-core/crypto"
 )
 
-var generatedPairs int64 = 0
+var globalSeed int64
 
 func RandTestKeyPair(typ, bits int) (ci.PrivKey, ci.PubKey, error) {
-	seed := time.Now().UnixNano()
-
 	// workaround for low time resolution
-	seed += atomic.AddInt64(&generatedPairs, 1) << 32
-
-	return SeededTestKeyPair(typ, bits, time.Now().UnixNano())
+	seed := atomic.AddInt64(&globalSeed, 1)
+	return SeededTestKeyPair(typ, bits, seed)
 }
 
 func SeededTestKeyPair(typ, bits int, seed int64) (ci.PrivKey, ci.PubKey, error) {
--- a/test/peer.go
+++ b/test/peer.go
@@ -1,10 +1,8 @@
 package test
 
 import (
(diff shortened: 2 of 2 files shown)


aschmahmann:
@Stebalien thanks for moving on this so quickly. Brief nitpick, I don't think this exactly matches up with suggestion 4 (remove using non-seeded randomness in tests) because the number (and order) of tests in a given execution run (and therefore the state of the PRNG) can vary.

However, this does fix the issue 👍.

Stebalien:
True, but it's as close as we can get without rewriting tests.

aschmahmann:
> True, but it's as close as we can get without rewriting tests.

Yep, fair enough. Not necessarily related to this PR do you have any thoughts on if/how we might want to rewrite the tests in the future or going forward? Should we create Seeded alternatives to the PeerID functions and recommend their use going forward?

I also wanted to throw out there that while it's kind of hacky if we had a global `SetSeed(int64)` function this would allow us to just put `SetSeed(0)` at the beginning of tests where we want predictable "random" numbers from test to test. However, it's worth noting that for certain cases (e.g. generating RSA keys) this approach won't even help us (as described in libp2p/go-libp2p-examples#20).

PR #45, opened 2019-07-30
Return error in Emit

bigs:
Could only break code depending on panic recovery. Proposed solution to https://github.com/libp2p/go-eventbus/issues/20.

### DIFF
--- a/event/bus.go
+++ b/event/bus.go
@@ -19,7 +19,7 @@ type Emitter interface {
 	// calls to Emit will block.
 	//
 	// Calling this function with wrong event type will cause a panic.
-	Emit(evt interface{})
+	Emit(evt interface{}) error
 }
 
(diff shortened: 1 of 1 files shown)


bigs:
Bump @magik6k. The `compat-check` fails for obvious reasons. Would this qualify a minor version release, or does it have to be major @raulk?

raulk:
Honestly, the eventbus is experimental, the change is backwards compatible **in practice** for consumers, and breaking for implementers, of which there’s only go-libp2p-eventbus. Call me a mischief, but I’d release the plaintext changeset as a minor, and this as a patch on top of that.

raulk:
Should be ok as long as we document this deviance in the release notes.

bigs:
sounds good @raulk

PR #46, opened 2019-08-01
cherry-pick key optimizations from stabilize

Stebalien:
These are all the non-controversial changes to crypto.


crypto/key.go
```diff
@@ -363,7 +364,21 @@ func KeyEqual(k1, k2 Key) bool {
...
-	b1, err1 := k1.Bytes()
-	b2, err2 := k2.Bytes()
-	return subtle.ConstantTimeCompare(b1, b2) == 1 && err1 == err2
+	return k1.Equals(k2)
+}
+
+func basicEquals(k1, k2 Key) bool {
+	if k1.Type() != k2.Type() {
+		return false
+	}
```
Stebalien: Note: This won't be constant time if the key _types_ differ but that's not private.

crypto/openssl_common.go
```diff
@@ -5,7 +5,7 @@ package crypto
 import (
 	pb "github.com/libp2p/go-libp2p-core/crypto/pb"
 
-	openssl "github.com/spacemonkeygo/openssl"
+	openssl "github.com/libp2p/go-openssl"
```
Stebalien: We forked this library as upstream isn't maintained. Our library adds:

* FreeBSD support.
* Key equality checking.

crypto/secp256k1.go
```diff
@@ -66,10 +66,10 @@ func (k *Secp256k1PrivateKey) Raw() ([]byte, error) {
 func (k *Secp256k1PrivateKey) Equals(o Key) bool {
 	sk, ok := o.(*Secp256k1PrivateKey)
 	if !ok {
-		return false
+		return basicEquals(k, o)
 	}
 
-	return k.D.Cmp(sk.D) == 0
+	return k.GetPublic().Equals(sk.GetPublic())
```
Stebalien: Bug! This is now constant time.

crypto/openssl_common.go
raulk: Added motivation for forking to the README.md in libp2p/go-openssl, for our future selves: https://github.com/libp2p/go-openssl/commit/1934e456a08650fa43aefb673eb660b7cac8c0b4

PR #50, opened 2019-08-14
Compare multihash `Code` to `IDENTITY` when extracting public key

ralexstokes:
This PR is a non-substantive change with the goal of making this method
`ExtractPublicKey` from a `peer.ID` clearer to the reader. This goal is
accomplished by using the full name of the identity hash code `IDENTITY` instead
of the deprecated `ID` code. This change makes its clearer to the reader that
there is a trivial way to compute the public key given the digest. Without this
change it is easy to confuse `ID` with a concept relating to peer IDs (and thus
implying this function somehow computes a hash pre-image to get the
corrresponding public key.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/minio/sha256-simd v0.1.0
 	github.com/mr-tron/base58 v1.1.2
 	github.com/multiformats/go-multiaddr v0.0.4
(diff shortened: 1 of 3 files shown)


ralexstokes:
@Stebalien thanks! i tried to use `v0.0.6` but the `go mod` tooling wasn't picking it up -- i saw you pushed `v0.0.7` which is working locally -- should pass the CI

PR #53, opened 2019-08-26
Add SecureMuxer interface to facilitate simultaneous open

vyzo:

Also fixes a bug in the insecure transport that was breaking tests in go-conn-security: the remote peer ID was not set for connections with `/plaintext/1.0.0`.

**Edit:**

The two main changes this PR introduces are:

1. A context API to instruct libp2p Swarm to force a direct connection to the peer even if a proxied connection to it already exists. This is to implemented the Hole Punching co-ordination Spec described in https://github.com/libp2p/specs/pull/173. You can see how this API is used at https://github.com/libp2p/go-libp2p/pull/711/files.

2. We introduce an interface that allows negotiating a security protocol between peers even if it's a simultaneous connect. This is for https://github.com/libp2p/specs/pull/196.


sec/security.go
```diff
@@ -24,3 +24,17 @@ type SecureTransport interface {
...
+	// SecureInbound secures an inbound connection.
+	// The returned boolean indicates whether the connection should be trated as a server
+	// connection; in the case of SecureInbound it should always be true.
+	SecureInbound(ctx context.Context, insecure net.Conn) (SecureConn, bool, error)
+
+	// SecureOutbound secures an outbound connection.
+	// The returned boolean indicates whether the connection should be treated as a server
+	// connection due to simultaneous open.
+	SecureOutbound(ctx context.Context, insecure net.Conn, p peer.ID) (SecureConn, bool, error)
+}
```
Stebalien: I'd like to consider using a special sentinel error instead. That is, return a connection _and_ a special `ErrSimultaniousConnect`.

This way we can avoid polluting the interfaces too much.

sec/security.go
```diff
@@ -24,3 +24,17 @@ type SecureTransport interface {
...
 	SecureOutbound(ctx context.Context, insecure net.Conn, p peer.ID) (SecureConn, error)
 }
+
+// A SecureMuxer is a wrapper around SecureTransport which can select security protocols
+// and open outbound connections with simultaneous open.
+type SecureMuxer interface {
+	// SecureInbound secures an inbound connection.
+	// The returned boolean indicates whether the connection should be trated as a server
+	// connection; in the case of SecureInbound it should always be true.
+	SecureInbound(ctx context.Context, insecure net.Conn) (SecureConn, bool, error)
```
Stebalien: Per our discussion, let's figure out _when_ this can happen. We normally won't get a "simultaneous accept".

sec/security.go
vyzo: What would the error buy is? We can still secure the connection with simultaneous open.

sec/security.go
vyzo: Well, there is symmetry in having the boolean indicator in the interface, plus it opens the door to active implementations that can deal with the situation.

sec/security.go
vyzo: also, simultaneous connect is _not_ an error!

Stebalien: I'm just trying to avoid having this TCP specific feature from leaking into the interface.

Stebalien: We discussed this synchronously. I was assuming we'd be able to come up with a better interface the next time we add a transport from scratch. However, @vyzo convinced me otherwise.

The only viable alternative is to add a third `Secure(ctx, insecure, p) (Conn, iAmInitiator bool, error)` function. However, that has no real benefit over `SecureOutbound`

Well, I guess it has a _slight_ benefit: Transports that _aren't_ TCP could skip this whole dance. Maybe _that's_ worth it @vyzo? Regardless, my hope to avoid messing with the interfaces seems moot.

sec/security.go
marten-seemann: @vyzo I have the same concern as @Stebalien. Can you explain what you mean by active implementations? Is that some kind of hole punching technique you're referring to?

vyzo: I don't have anything that is doing active listen yet, but we could conceivably solve the accept-accept scenario in hole punching (that according to the literature _can_ happen, we'll see in practice).
Other than that, it's symmetry of the interface.

marten-seemann: What is the accept-accept scenario? Opening a TCP socket is a purely local operation that doesn't send any packets on the wire, isn't it? So this can't possibly lead to the establishment of a connection.

vyzo: It something that can happen with hole-punching, because you are both listening and opening a connection on the same time.
So it is conceivable that the `SYN`s will get through the NAT/firewall and the OS routes them to the listening socket instead of the connecting socket.
It is supposed to be unlikely, but something that can happen nonetheless.

aarshkshah1992:
@Stebalien @marten-seemann @vyzo 

Given our goal of getting to TCP hole punching quickly, I'd like to put a hold on the discussion and use this interface changes as is because it does do a good job of solving the simultaneous connect problem in `go-multistream`.

I am happy to mark the interface as experimental and working off this feature branch instead of merging to master if that helps us make progress for now.

Please let me know if there are strong objections to this,

aarshkshah1992:
Hey, his was subsusmed in #180 and #181 . **However,**  I believe the _"Also fixes a bug in the insecure transport that was breaking tests in go-conn-security: the remote peer ID was not set for connections with /plaintext/1.0.0."_ stuff hasn't made in I think. We can probably raise a separate PR just for that and get it in.

vyzo:
yeah, let's cherry-pick that.

PR #56, opened 2019-09-04
Fix documentation missmatch for crypto methods.

JorikSchellekens:
The docs seem to lie about what the `ConfigEncodeKey` and `ConfigDecodeKey` methods do. Currently they seem to simply be wrappers around base 64 encoding and decoding.

The simple fix is to update the docs but is it possible to get these methods
to marshal to public or private key correctly by passing a "type" (reflect.Type or something) to them?

(First day using go, otherwise I would have taken an actual stab at this)

### DIFF
--- a/crypto/key.go
+++ b/crypto/key.go
@@ -349,12 +349,12 @@ func MarshalPrivateKey(k PrivKey) ([]byte, error) {
 	return proto.Marshal(pbmes)
 }
 
-// ConfigDecodeKey decodes from b64 (for config file), and unmarshals.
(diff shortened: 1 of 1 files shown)


JorikSchellekens:
This is not covered by #2

JorikSchellekens:
Also, I've no idea what the codecov is complaining about. This is a doc only change so far.

JorikSchellekens:
Would it make more sense to just have a methods for both PrivateKeys and PublicKeys?

PR #59, opened 2019-09-19
adds ability to create keys from crypto.PrivateKey types

adam-hanna:


crypto/key.go
```diff
@@ -174,6 +180,42 @@ func GenerateEKeyPair(curveName string) ([]byte, GenSharedKey, error) {
 	return pubKey, done, nil
 }
 
+// KeyPairFromKey generates a new private and public key from an input private key
+func KeyPairFromKey(priv crypto.PrivateKey) (PrivKey, PubKey, error) {
+	if priv == nil {
+		return nil, nil, ErrNilPrivateKey
+	}
+
+	switch priv.(type) {
```
Stebalien: use `switch priv := priv.(type)` to save the result. That way, you don't need to cast in all the cases.

crypto/key.go
```diff
@@ -174,6 +180,42 @@ func GenerateEKeyPair(curveName string) ([]byte, GenSharedKey, error) {
...
+	case *rsa.PrivateKey:
+		log.Println("1")
```
Stebalien: debugging code.

crypto/key.go
```diff
@@ -174,6 +180,42 @@ func GenerateEKeyPair(curveName string) ([]byte, GenSharedKey, error) {
 	return pubKey, done, nil
 }
 
+// KeyPairFromKey generates a new private and public key from an input private key
+func KeyPairFromKey(priv crypto.PrivateKey) (PrivKey, PubKey, error) {
```
Stebalien: This will need to work when `-tags=openssl` is passed.

crypto/key.go
adam-hanna: ack, embarrassing!

crypto/key.go
adam-hanna: Done

crypto/key.go
adam-hanna: done

adam-hanna:
Thanks for the help!

adam-hanna:
Scratch this, this isn't working...

`// +build !openssl` makes my code unable to find `KeyPairFromKey`

Stebalien:
@raulk this extends the core interface slightly but seems useful (and looks correct).

crypto/key_not_openssl.go
```diff
@@ -0,0 +1,40 @@
...
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rsa"
+
+	btcec "github.com/btcsuite/btcd/btcec"
+	"golang.org/x/crypto/ed25519"
+)
+
+// KeyPairFromKey generates a new private and public key from an input private key
+func KeyPairFromKey(priv crypto.PrivateKey) (PrivKey, PubKey, error) {
```
raulk: How about we name this `KeyPairFromStdKey` to more accurately describe what’s going on?

crypto/key_not_openssl.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rsa"
+
+	btcec "github.com/btcsuite/btcd/btcec"
+	"golang.org/x/crypto/ed25519"
+)
+
+// KeyPairFromKey generates a new private and public key from an input private key
```
raulk: Godoc is a bit bleak. “Generates” is an imprecise term here, as there’s no key generation happening, rather wrapping or conversion.

crypto/key_not_openssl.go
adam-hanna: done

crypto/key_not_openssl.go
adam-hanna: done

adam-hanna:
Thanks @raulk 

Done!

Stebalien:
Thanks!

PR #60, opened 2019-09-24
crypto: improve key comparison logic

Stebalien:


Stebalien:
> what is basicEquals?

It's the fallback method for comparing two keys when we can't special-case it. The special-cased versions all try to avoid serializing the keys for better performance.

PR #63, opened 2019-10-01
Remove Public Key

aarshkshah1992:
For https://github.com/libp2p/go-libp2p-peerstore/issues/93
Looks like only Remove Public Key is required.


aarshkshah1992:
@raulk  Apologies. Have linked the issue in the description now. This is to enables us to remove the public key from the peerstore when a peer disconnects so as to not cause memory leaks.

Implementation is at https://github.com/libp2p/go-libp2p-peerstore/pull/94

raulk:
Functionality landed in https://github.com/libp2p/go-libp2p-peerstore/pull/184/files.

PR #73, opened 2019-11-08
Signed envelopes & routing records

yusefnapora:
This implements the signed envelopes and routing records described in https://github.com/libp2p/specs/pull/217.

I've also updated the peerstore implementation to fit the new API & will open a PR there in a sec.

Travis correctly marks this as a breaking API change, since the new methods on the peerstore make existing implementations no longer fit the interface. I think this means I need to bump the minor version?

I still feel like a Go newbie, so don't be shy about review feedback :)

Some TODO items:

- [x] filter out local addresses when making a routing record for a Host
- [x] ensure envelope type hint has valid multicodec prefix (right now it accepts any byte string)
  - maybe? we'd have to keep this updated as new multicodecs are added


crypto/envelope.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (e *SignedEnvelope) Equals(other *SignedEnvelope) bool {
+	return e.PublicKey.Equals(other.PublicKey) &&
+		bytes.Compare(e.TypeHint, other.TypeHint) == 0 &&
+		bytes.Compare(e.contents, other.contents) == 0 &&
+		bytes.Compare(e.signature, other.signature) == 0
+}
+
+// makeSigBuffer is a helper function that prepares a buffer to sign or verify.
+func makeSigBuffer(domain string, typeHint []byte, content []byte) []byte {
```
Stebalien: Nit:

1. It would be nice to use the buffer pool (go-buffer-pool) here.
2. We can probably do this in one large over-allocation.

crypto/envelope.go
```diff
@@ -0,0 +1,139 @@
...
+		PublicKey: privateKey.GetPublic(),
+		TypeHint:  typeHint,
+		contents:  contents,
+		signature: sig,
+	}, nil
+}
+
+// UnmarshalEnvelope converts a serialized protobuf representation of an envelope
+// into a SignedEnvelope struct.
+func UnmarshalEnvelope(serializedEnvelope []byte) (*SignedEnvelope, error) {
```
Stebalien: Two step signature verification processes where the user can forget the second step make me nervous. How about:

* `UnmarshalWithoutValidating(bytes) (*SignedEnvelope, error)`
* `Open(bytes, domain) (*SignedEnvelope, error)`

That way, we can't end up with an unvalidated `SignedEnvelope` object unless the user very explicitly opts in.

crypto/envelope.go
```diff
@@ -0,0 +1,139 @@
...
+		PublicKey: key,
+		TypeHint:  e.TypeHint,
+		contents:  e.Contents,
+		signature: e.Signature,
+	}, nil
+}
+
+// Validate returns true if the envelope signature is valid for the given 'domain',
+// or false if it is invalid. May return an error if signature validation fails.
+func (e *SignedEnvelope) Validate(domain string) (bool, error) {
```
Stebalien: Along with my above comment, I'd remove this method. Basically, make users choose to either validate or not validate up-front.

crypto/pb/crypto.proto
```diff
@@ -18,3 +18,10 @@ message PrivateKey {
 	required KeyType Type = 1;
 	required bytes Data = 2;
 }
+
+message SignedEnvelope {
+	required PublicKey PublicKey = 1;
+	required bytes TypeHint = 2;
```
Stebalien: Required?

routing/pb/routing_state.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
```
Stebalien: We should probably be moving to proto3.

crypto/envelope.go
yusefnapora: I like this - I tried to force validation by making the payload private & requiring you to "open" the envelope (and validate) to get the payload. But you could still end up passing on an invalid envelope to someone else if you never open the payload.

Should we also be validating the signature before we marshal to bytes? Then if someone invalidates the signature by changing e.g. the `payloadType` field, they won't be able to write the record out. We'd need the `Marshal` method to take the domain string in order to validate. It might also be worth having an unsafe `MarshalWithoutValidating` method for testing or if you explicitly want to opt out of validation.

Stebalien: > Should we also be validating the signature before we marshal to bytes?

Meh. I'm fine with marshaling without verifying. We could also just make all the fields private but that may not be worth it. I don't really have any strong opinions here.

I also didn't realize that the payload was private. That resolves my initial concern (although the type is really a part of the payload and should probably also be private). However, I think validating when deserializing is likely going to be more efficient in general and make it easier to pass around signed envelopes.

yusefnapora: Hmm, maybe instead of validating before marshaling, we can just make all the fields private, and expose read-only accessor methods, so it's not possible to alter the struct contents after you make or open an envelope?

crypto/envelope.go
```diff
@@ -0,0 +1,191 @@
+package crypto
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"github.com/golang/protobuf/proto"
```
vyzo: we use gogo protobuf, not the google-vendored one.

crypto/envelope.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, err
+	}
+	return e, nil
+}
+
+// UnmarshalEnvelopeWithoutValidating unmarshals a serialized SignedEnvelope protobuf message,
+// without validating its contents. Should not be used unless you have a very good reason
+// (e.g. testing)!
+func UnmarshalEnvelopeWithoutValidating(serializedEnvelope []byte) (*SignedEnvelope, error) {
+	e := pb.SignedEnvelope{}
```
vyzo: `var e pb.SignedEnvelope` is more idiomatic.

crypto/envelope.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// A binary identifier that indicates what kind of data is contained in the payload.
+	// TODO(yusef): enforce multicodec prefix
+	payloadType []byte
+
+	// The envelope payload.
+	payload []byte
+
+	// The signature of the domain string, type hint, and payload.
+	signature []byte
```
vyzo: Why not make the fields public? This way we can avoid the accessor functions and use it as a proper struct.

crypto/pb/envelope.proto
```diff
@@ -0,0 +1,12 @@
...
+package crypto.pb;
+
+import "crypto.proto";
+
+message SignedEnvelope {
+    PublicKey public_key = 1;
+    bytes payload_type = 2;
+    bytes payload = 3;
+    bytes signature = 4;
+}
```
vyzo: newline!

crypto/envelope.go
```diff
@@ -0,0 +1,191 @@
+package crypto
+
+import (
```
vyzo: A small thing, but as a convention, we usually group our imports -- stdlib imports go first, unnamed imports then, and named imports last.

peerstore/peerstore.go
```diff
@@ -120,6 +129,18 @@ type AddrBook interface {
...
+	// given peer id, if one exists in the peerstore. The record is
+	// returned as a byte slice containing a serialized SignedEnvelope.
+	// Returns nil if no routing state exists for the peer.
+	SignedRoutingState(p peer.ID) []byte
+
+	// SignedRoutingStates returns signed RoutingState records for each of
+	// the given peer ids, if one exists in the peerstore.
+	// Returns a map of peer ids to serialized SignedEnvelope messages. If
+	// no routing state exists for a peer, their map entry will be nil.
+	SignedRoutingStates(peers ...peer.ID) map[peer.ID][]byte
```
vyzo: I think that's an overkill, why do we need this method? It bloats the interface somewhat.

routing/pb/routing_state.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+package routing.pb;
+
+message RoutingStateRecord {
+    message AddressInfo {
+        bytes multiaddr = 1;
+    }
+
+    bytes peer_id = 1;
```
vyzo: we usually camelCase those.

routing/state.go
```diff
@@ -0,0 +1,181 @@
...
+// The type hint used to identify routing state records in a SignedEnvelope.
+// TODO: register multicodec
+var StateEnvelopePayloadType = []byte("/libp2p/routing-state-record")
+
+// AnnotatedAddr will extend the Multiaddr type with additional metadata, as
+// extensions are added to the routing state record spec. It's defined now to
+// make refactoring simpler in the future.
+type AnnotatedAddr struct {
+	ma.Multiaddr
+}
```
vyzo: This seems redundant for now; adding fields to the struct is a breaking change and it makes it awkward to work with.
So  if we ever introduce annotations to the multiaddr, we can make the annotated addr struct just in time.

routing/state.go
```diff
@@ -0,0 +1,181 @@
...
+// Note that Seq values are peer-specific and can only be compared for records with equal PeerIDs.
+type RoutingState struct {
+	// PeerID is the ID of the peer this record pertains to.
+	PeerID peer.ID
+
+	// Seq is an increment-only sequence counter used to order RoutingState records in time.
+	Seq uint64
+
+	// Addresses contains the public addresses of the peer this record pertains to.
+	Addresses []*AnnotatedAddr
```
vyzo: that's awkward to work with.

routing/state.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// RoutingStateWithMultiaddrs returns a RoutingState record for the given peer id
+// that contains the given multiaddrs. It generates a timestamp-based sequence number.
+func RoutingStateWithMultiaddrs(p peer.ID, addrs []ma.Multiaddr) *RoutingState {
+	annotated := make([]*AnnotatedAddr, len(addrs))
+	for i, a := range addrs {
+		annotated[i] = &AnnotatedAddr{Multiaddr: a}
+	}
```
vyzo: so awkward!

routing/state.go
```diff
@@ -0,0 +1,181 @@
...
+	return proto.Marshal(&msg)
+}
+
+// Multiaddrs returns the addresses from a RoutingState record without any metadata annotations.
+func (s *RoutingState) Multiaddrs() []ma.Multiaddr {
+	out := make([]ma.Multiaddr, len(s.Addresses))
+	for i, addr := range s.Addresses {
+		out[i] = addr.Multiaddr
+	}
```
vyzo: Someone is bound to scream at you for this :)
The idiomatic way to do it is
```
out := make([]ma.Multiaddr, 0, len(s.Addresses)
for _, addr := range s.Addresses {
  out = append(out, addr)
}
```

crypto/envelope.go
yusefnapora: after sync discussion with @vyzo & @Stebalien we decided to just expose the fields - if you mutate them it's your own fault when validation fails 😄

routing/pb/routing_state.proto
yusefnapora: I noticed that, but someone (maybe @raulk; I can't seem to find the original comment now) mentioned that snake_case is [recommended by google](https://developers.google.com/protocol-buffers/docs/style#message-and-field-names) - it gets mapped to camelCase in the generated code (except langs like c++ and maybe python).

I don't really care either way, but it does bug me that I'm adding code that isn't consistent with the rest of our protos... if we plan to change everything to snake_case though, I'd rather start doing it now.

yusefnapora: found the comment I was referring to above; it's from the RFC: https://github.com/libp2p/specs/pull/217#discussion_r345124996

vyzo: all our protobufs use camelCase though.

yusefnapora: yeah. I think I'll change it back to camelCase & leave it as snake case in the RFC. if we decide we want go-libp2p to use snake case, we should convert them all in one pass, so they stay consistent.

yusefnapora:
I've been playing around with interfaces a bit, and there is a way to add this functionality without making a breaking API change.

Instead of adding the new methods to the `peerstore.AddrBook` interface, we can define a new `CertifiedAddrBook` interface that just contains the new methods (`AddCertifiedAddrs`, `CertifiedAddrs` and `SignedRoutingState`).

To avoid the breaking change, we can't embed the new interface into the main `Peerstore` interface though, so callers would have to type assert to check if the peerstore supports signed addresses, e.g.

```go
p := myHost.Peerstore()
if cab, ok := p.(CertifiedAddrBook); ok {
  cab.AddCertifiedAddrs(signedRoutingState, ttl)
}
```

This is more awkward to use though, and I'm probably overthinking this...

vyzo:
Definitely overthinking this! I think it's ok to just extend the interface in this case.

host/helpers.go
```diff
@@ -9,3 +13,14 @@ func InfoFromHost(h Host) *peer.AddrInfo {
 		Addrs: h.Addrs(),
 	}
 }
+
+// SignedRoutingStateFromHost returns a SignedRoutingState record containing
+// the Host's listen addresses, signed with the Host's private key.
+func SignedRoutingStateFromHost(h Host) (*routing.SignedRoutingState, error) {
```
yusefnapora: This seems like the place to filter out LAN addresses. I found [WANSharableAddrs in addrutil](https://godoc.org/github.com/libp2p/go-addr-util#WANShareableAddrs), which seems like it should do the trick.

We want to be able to opt-in to including LAN addresses though, and I'm not sure what the best way to do that is, since this is a free function. I guess we could use a global variable, but that seems messy.

Maybe this function should be a method on the `Host` instead, and we can set a flag when constructing the Host instance in go-libp2p? @vyzo, thoughts?

routing/state.go
```diff
@@ -0,0 +1,161 @@
...
+	return SignedRoutingStateFromEnvelope(envelope)
+}
+
+// SignedRoutingStateFromEnvelope accepts a SignedEnvelope struct containing
+// a RoutingStateRecord protobuf and returns a SignedRoutingState record.
+// Fails if the signature is invalid, if the envelope has an unexpected payload type,
+// or if deserialization of the envelope payload fails.
+func SignedRoutingStateFromEnvelope(envelope *crypto.SignedEnvelope) (*SignedRoutingState, error) {
+	if bytes.Compare(envelope.PayloadType, StateEnvelopePayloadType) != 0 {
```
yusefnapora: I realized this is probably too strict, since we might want to change the payload_type later if we register a new multicodec code for routing records. We could log a warning, but I don't think go-libp2p-core has a logging framework, so maybe it's better for this method to just ignore the payload type. Unmarshal will fail anyway if the payload is actually the wrong type.

host/helpers.go
yusefnapora: BTW, the reason I didn't make this a method on Host in the first place was just to avoid breaking the Host interface. I know we have a few implementations (basichost and the ones in go-libp2p), but maybe downstream users have their own that we don't know about?

vyzo: It can take a boolean for the option, or utilize the functional options pattern.
I doubt anyone else has a host implementation, but it is fine with this being free-standing.

yusefnapora: hmm, okay, it looks like [we want to deprecate go-addr-util](https://github.com/libp2p/go-addr-util/issues/18) and some digging reveals that it doesn't filter out standard LAN subnets, just loopback and IPv6 link-local addrs. So perhaps the filtering will take a little more work.

vyzo: Check out go-multiaddr-net; we have an `IsPublicAddr` utility function you could use for the filter.

yusefnapora: thanks, that works 👍

host/helpers.go
```diff
@@ -9,3 +16,54 @@ func InfoFromHost(h Host) *peer.AddrInfo {
...
+
+// IncludeLocalAddrs can be passed into SignedRoutingStateFromHost to
+// produce a routing record with LAN and loopback addresses included.
+func IncludeLocalAddrs(cfg *config) {
+	cfg.includeLocalAddrs = true
+}
+
+// minimalHost is the subset of the Host interface that's required by
+// SignedRoutingStateFromHost.
+type minimalHost interface {
```
yusefnapora: I added this to make testing a bit simpler, but can remove it if it makes things more confusing. I still ended up having to define a ton of stubs to mock out the `Peerstore`, so it doesn't really help that much 😛

crypto/envelope.go
```diff
@@ -0,0 +1,171 @@
...
+	PayloadType []byte
+
+	// The envelope payload.
+	Payload []byte
+
+	// The signature of the domain string, type hint, and payload.
+	signature []byte
+}
+
+var errEmptyDomain = errors.New("envelope domain must not be empty")
```
raulk: Let's export these errors.

routing/state.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/gogo/protobuf/proto"
+	"github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/peer"
+	pb "github.com/libp2p/go-libp2p-core/routing/pb"
+	ma "github.com/multiformats/go-multiaddr"
+	"time"
+)
+
+// The domain string used for routing state records contained in a SignedEnvelope.
+const StateEnvelopeDomain = "libp2p-routing-state"
```
raulk: I think routing is a bad name for this. Routing is _a process_ for which a record of this kind _can_ be used. But it's not the only one. I'd rather call this a "signed peer record", and put it in the peer package.

routing/state.go
```diff
@@ -0,0 +1,157 @@
...
+// The type hint used to identify routing state records in a SignedEnvelope.
+// TODO: register multicodec
+var StateEnvelopePayloadType = []byte("/libp2p/routing-state-record")
+
+type SignedRoutingState struct {
+	// PeerID is the ID of the peer this record pertains to.
+	PeerID peer.ID
+
+	// Seq is an increment-only sequence counter used to order RoutingState records in time.
+	Seq uint64
```
raulk: Once again, why is the seq _inside_ the payload? I don't regard this as part of the payload itself, and it's very likely that all records need version management, so it makes sense to move it to the outer layer.

If it's for signing purposes, we can concatenate the sequence number to the payload plaintext prior to signing.

routing/state.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Addrs contains the public addresses of the peer this record pertains to.
+	Addrs []ma.Multiaddr
+
+	// Envelope contains the signature and serialized RoutingStateRecord protobuf.
+	// Although it uses a bit
+	Envelope *crypto.SignedEnvelope
```
raulk: This is weird. The envelope should wrap the payload, not the other way around.

raulk:
> ensure envelope type hint has valid multicodec prefix (right now it accepts any byte string)
>  * maybe? we'd have to keep this updated as new multicodecs are added

Yeah, I wouldn't hardcode an enumeration in the protobuf.

raulk:
> This is more awkward to use though, and I'm probably overthinking this...

Note that the using component would only do that once, and it would safe a reference to the `CertifiedAddrBook`.

routing/state.go
yusefnapora: Sounds good to me 👍

routing/state.go
yusefnapora: Keeping the original envelope around lets us marshal / unmarshal it without invalidating the signature.  Essentially `SignedRoutingState` is a composition of a `SignedEnvelope` with an unpacked representation of the payload.

I agree this is weird though, and it looks like I got distracted and didn't finish writing the comment to explain it 😄 .

I'm pretty sure the reason I ended up with this was to enable an api like:

```go
type CertifiedAddrBook interface {
  AddCertifiedAddrs(record *SignedRoutingState)
}
```

instead of `AddCertifiedAddrs(envelope *SignedEnvelope)`, where `AddCertifiedAddrs` would be responsible for "opening" the envelope. 

Although looking at it now, that might actually be easier to use, since the caller of `AddCertifiedAddrs` wouldn't have to open the envelope first...

routing/state.go
yusefnapora: I think @Stebalien's objection was that it makes the envelope less generic. I think I'd prefer to have it on the outer envelope also, since it would let you quickly reject old records without having to deserialize the payload.

yusefnapora:
Hey @raulk, I pulled in your review branch and added a few commits based on your feedback.

The main changes are:

- Defined a separate `CertifiedAddrBook` and a helper method `GetCertifiedAddrBook(AddrBook) (CertifiedAddrBook, bool)` to do the upcasting.
  - yay for API compatibility & the elusive green checkmark :)
- Renamed the record from `RoutingStateRecord` to `PeerRecord` and moved to the `peer` package.
- Removed the embedded reference to the `SignedEnvelope` from the `PeerRecord`. It's now just a representation of the envelope payload, with helpers to produce a `SignedEnvelope` and unmarshal a `PeerRecord` from a `SignedEnvelope`.
- The `AddCertifiedAddrs` method on `CertifiedAddrBook` now takes a generic `SignedEnvelope` as an argument, and unpacks the `PeerRecord` payload. Likewise, `CertifiedAddrBook.SignedPeerRecord(peerId)` returns a `SignedEnvelope`.
- Moved the `Seq` field from the record to the `SignedEnvelope` and added to the signature calculation
  - will need to update the spec


What do you think? I've updated my -peerstore and go-libp2p branches locally to reflect the changes & will push after some cleanup if you think we're on the right track here.

peer/pb/peer_record.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+package peer.pb;
+
+message PeerRecord {
+    message AddressInfo {
+        bytes multiaddr = 1;
+    }
+
+    bytes peerId = 1;
```
raulk: (bikeshed alert) Pretty please, can we use Protobuf style conventions and name fields with snake case? https://developers.google.com/protocol-buffers/docs/style#message-and-field-names

libp2p has upper camel, lower camel, and snake casing across the codebase, and the correct one is the latter :-) Fortunately, fixing the casing in other protobufs should not break anything.

peer/record.go
```diff
@@ -0,0 +1,188 @@
+package peer
+
+import (
+	"errors"
+
+	"github.com/gogo/protobuf/proto"
```
raulk: Group all libp2p imports first, then all third-party libraries in another group.

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+	pb "github.com/libp2p/go-libp2p-core/peer/pb"
+	"github.com/libp2p/go-libp2p-core/record"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// The domain string used for peer records contained in a SignedEnvelope.
+const PeerRecordEnvelopeDomain = "libp2p-peer-record"
+
+// The type hint used to identify peer records in a SignedEnvelope.
+// TODO: register multicodec
```
raulk: Do we have a top-level tracking issue with all TODOs and PRs for this endeavour? If not, we totally should!

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+var PeerRecordEnvelopePayloadType = []byte("/libp2p/peer-record")
+
+// ErrPeerIdMismatch is returned when attempting to sign a PeerRecord using a key that
+// does not match the PeerID contained in the record.
+var ErrPeerIdMismatch = errors.New("unable to sign peer record: key does not match record.PeerID")
```
raulk:
```suggestion
var ErrPeerIDMismatch = errors.New("unable to sign peer record: key does not match record.PeerID")
```

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+
+// PeerRecord contains information that is broadly useful to share with other peers,
+// either through a direct exchange (as in the libp2p identify protocol), or through
+// a Peer Routing provider, such as a DHT.
+//
+// Currently, a PeerRecord contains the public listen addresses for a peer, but this
+// is expected to expand to include other information in the future.
+//
+// PeerRecords are intended to be signed by the peer they describe, and there are no
```
raulk:
```suggestion
// PeerRecords are intended to be self-signed by the peer they describe, and there are no
```

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+// public methods for marshalling and unmarshalling unsigned PeerRecords.
+//
+// To share a PeerRecord, first call Sign to wrap the record in a SignedEnvelope
+// and sign it with the local peer's private key:
+//
+//     rec := NewPeerRecord(myPeerId, myAddrs)
+//     envelope, err := rec.Sign(myPrivateKey)
```
raulk: Aha, that's interesting. If `Sign()` is a member of the record type, wouldn't it have to be duplicated everywhere? A more ergonomic API is to hoist the `Sign` method to the `record` package, and call it something like `Seal`, which returns an envelope. Really, signing a record is one operation that is part of the larger act of authenticating and securing a record.

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+//
+// The resulting record.SignedEnvelope can be marshalled to a []byte and shared
+// publicly. As a convenience, the MarshalSigned method will produce the
+// SignedEnvelope and marshal it to a []byte in one go:
+//
+//     rec := NewPeerRecord(myPeerId, myAddrs)
+//     recordBytes, err := rec.MarshalSigned(myPrivateKey)
```
raulk: Following from the above, we can call this `record.SealAndMarshal()`.

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+// and sign it with the local peer's private key:
+//
+//     rec := NewPeerRecord(myPeerId, myAddrs)
+//     envelope, err := rec.Sign(myPrivateKey)
+//
+// The resulting record.SignedEnvelope can be marshalled to a []byte and shared
+// publicly. As a convenience, the MarshalSigned method will produce the
+// SignedEnvelope and marshal it to a []byte in one go:
+//
+//     rec := NewPeerRecord(myPeerId, myAddrs)
```
raulk: I'm afraid that `NewPeerRecord` will break frequently as we add more items to the peer record. Perhaps not providing a constructor, and having the consumer populate the struct directly is more future-proof?

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+//     recordBytes, err := rec.MarshalSigned(myPrivateKey)
+//
+// To validate and unmarshal a signed PeerRecord from a remote peer, use the
+// UnmarshalSignedPeerRecord function:
+//
+//    rec, envelope, err := UnmarshalSignedPeerRecord(recordBytes)
```
raulk: Usually in go, you unmarshal into an existing (empty) struct. This allows you to control memory allocation and potentially reuse structs.

So this would look like:

```go
var rec *PeerRecord
err := rec.Unmarshal(bytes)
```

I also think that unsealing the envelope should return the bytes, and the type of the record. And the caller should be responsible for unmarshalling the record into the concrete backing struct.

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+	// PeerID is the ID of the peer this record pertains to.
+	PeerID ID
+
+	// Addrs contains the public addresses of the peer this record pertains to.
+	Addrs []ma.Multiaddr
+}
+
+// NewPeerRecord creates a PeerRecord with the given ID and addresses.
+// It generates a timestamp-based sequence number.
+func NewPeerRecord(id ID, addrs []ma.Multiaddr) *PeerRecord {
```
raulk: As I pointed out above, this constructor provides little value and is subject to frequent breakage. I'd kill it.

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+	return &PeerRecord{
+		PeerID: id,
+		Addrs:  addrs,
+	}
+}
+
+// PeerRecordFromAddrInfo creates a PeerRecord from an AddrInfo struct.
+// It generates a timestamp-based sequence number.
+func PeerRecordFromAddrInfo(info AddrInfo) *PeerRecord {
```
raulk: Contrary to be above, this constructor is OK, because it clearly bounds the input now and forever.

peerstore/peerstore.go
```diff
@@ -5,6 +5,7 @@ package peerstore
 import (
 	"context"
 	"errors"
+	"github.com/libp2p/go-libp2p-core/record"
```
raulk: Move import to the libp2p import group.

event/addrs.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/libp2p/go-libp2p-core/record"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// EvtLocalAddressesUpdated should be emitted when the set of listen addresses for
+// the local host changes. This may happen for a number of reasons. For example,
+// we may have opened a new relay connection, established a new NAT mapping via
+// UPnP, or been informed of our observed address by another peer.
+type EvtLocalAddressesUpdated struct {
```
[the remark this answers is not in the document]
yusefnapora: I'm glad you called this out, because I really wasn't sure how to best approach this.

I think you're right that full snapshots are better than these diffs; in fact, I had to awkwardly keep a reference to the Host in the consumer code in go-libp2p, since it needs the full current state to produce the PeerRecord, so the diff was being produced and then ignored...

I like option four, maybe with the removed addresses in a separate list, which would let you ignore them completely if you just want current addrs:

```
type AddrAction int
const (
  Unknown AddrAction = iota
  Added
  Maintained
  Removed
)

type UpdatedAddress struct {
  ma.Multiaddr
  Action AddrAction
}

type EvtLocalAddressesUpdated struct {
  CurrentAddrs []UpdatedAddress
  Removed []UpdatedAddress
}
```

The above uses an `Unknown` hint for producers that only have the current state and can't produce the diff hints. With that, do you think we still need a flag for whether hints are available?

peer/record.go
yusefnapora: Makes sense 👍

peer/record.go
yusefnapora: Sounds good 👍

Looking at it now, I agree that this does too much, and we should let the caller deal with unwrapping the envelope. I'll change this to accept an empty `*PeerRecord`.

event/addrs.go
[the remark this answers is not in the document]
raulk: That implementation proposal is good. My gut feeling tells me to replace the slices with `map[ma.Multiaddr]AddrAction`, so that the consumer can index by known multiaddrs, but that's worthless because `ma.Multiaddr` is actually an interface and therefore what's stored in the key is a pointer, which will be different every time. Indexing by a pointer in this case is useless.

We might still need to know if the publisher is sending hints. Otherwise, we can't disambiguate between the publisher not tracking at all, or `Removed` just being empty in this case because no addresses were removed. WDYT?

record/envelope.go
```diff
@@ -0,0 +1,196 @@
...
+type SignedEnvelope struct {
+	// The public key that can be used to verify the signature and derive the peer id of the signer.
+	PublicKey crypto.PubKey
+
+	// A binary identifier that indicates what kind of data is contained in the payload.
+	// TODO(yusef): enforce multicodec prefix
+	PayloadType []byte
+
+	// A monotonically-increasing sequence counter for ordering SignedEnvelopes in time.
+	Seq uint64
```
raulk: Now we have a sequence number here and inside the peer record?

event/addrs.go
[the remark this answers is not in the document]
yusefnapora: That makes sense - the flag would be useful for the `Removed` case.

We could do a map of hints if we used the `string(addr.Bytes())` as the keys and added a helper method to the event type. So the caller could do e.g. `evt.GetAction(addr)`. Then we wouldn't need the `UpdatedAddress` type, which would be nice for consumers that don't care about the hints at all.

peer/record.go
```diff
@@ -0,0 +1,188 @@
...
+	return NewPeerRecord(info.ID, info.Addrs)
+}
+
+// UnmarshalSignedPeerRecord accepts a []byte containing a SignedEnvelope protobuf message.
+// It will try to validate the envelope signature and unmarshal the payload as a PeerRecord.
+// Returns the PeerRecord and the SignedEnvelope if successful.
+func UnmarshalSignedPeerRecord(envelopeBytes []byte) (*PeerRecord, *record.SignedEnvelope, error) {
```
raulk: So here's what I've been thinking.

1. The `record` package defines an `Record` interface with two methods: `MarshalRecord()` and `UnmarshalRecord()`.
2. Record types such as `PeerRecord` implement that interface.
3. The `record` package exposes a global `RegisterType(mc byte[], typ interface{})` function. This creates a mapping between the multicodec and the type. Packages providing record types call this function in their `init()` func.
4. The `ConsumeEnvelope` function returns an extra out param `interface{}`: the unmarshalled record. We use the envelope's record type and reflection to unmarshal the payload.
5. For the cases where we don't want to use reflection, and we expect a particular type, we can have a `ConsumeTypedEnvelope` function that _takes_ a pointer to the struct to unmarshal the payload into.

yusefnapora: Ah, I like that!

I keep messing around trying to make something like `Envelope<T>` happen, where `T` is the payload type. I think your proposal is as close as we can get in go :)

Also, I just learned that packages can have an `init` function, so that's cool 😄

event/addrs.go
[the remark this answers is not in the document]
raulk: That would work, but in all honesty, it's sounding a bit convoluted at this point. I'd just add a third `Diffs bool` field:

```go
type AddrAction int

const (
  Unknown AddrAction = iota
  Added
  Maintained
  Removed
)

type UpdatedAddress struct {
  Address ma.Multiaddr
  Action AddrAction
}

type EvtLocalAddressesUpdated struct {
  Diffs        bool
  Current      []UpdatedAddress
  Removed      []UpdatedAddress
}
```

record/envelope.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/multiformats/go-varint"
+)
+
+// SignedEnvelope contains an arbitrary []byte payload, signed by a libp2p peer.
+//
+// Envelopes are signed in the context of a particular "domain", which is a
+// string specified when creating and verifying the envelope. You must know the
+// domain string used to produce the envelope in order to verify the signature
+// and access the payload.
+type SignedEnvelope struct {
```
raulk: Let's just simplify and call this `Envelope`. `SignedEnvelope` makes the assumption that this is the only thing we'll ever do with an envelope.

record/envelope.go
raulk: Let's keep it inside the peer record.

record/envelope.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// A monotonically-increasing sequence counter for ordering SignedEnvelopes in time.
+	Seq uint64
+
+	// The envelope payload.
+	Payload []byte
+
+	// The signature of the domain string :: type hint :: payload.
+	signature []byte
+}
```
raulk: Can we add a `cached record.Record` field where we stash the unmarshalled payload? We should also add an accessor `Record() Record`. Rationale: we might want to parse and validate an envelope in some layer, then pass the envelope to another layer (e.g. peerstore), omitting the cost of re-unmarshalling.

peerstore/peerstore.go
```diff
@@ -122,6 +123,66 @@ type AddrBook interface {
...
+	// addresses in the CertifiedAddrBook, the new addresses will replace the
+	// older ones, iff the new record has a higher sequence number than the
+	// existing record. Attempting to add a peer record with a
+	// sequence number that's <= an existing record for the same peer will not
+	// result in an error, but the record will be ignored.
+	//
+	// If the CertifiedAddrBook is also an AddrBook (which is most likely the case),
+	// adding certified addresses for a peer will *replace* any
+	// existing non-certified addresses for that peer, and only the certified
+	// addresses will be returned from AddrBook.Addrs.
```
raulk:
```suggestion
	// addresses will be returned from AddrBook.Addrs thereafter.
```

raulk: This is a great amount of detail. Awesome! Just to not miss the forest for the trees, I'd recap the motivation/rational for this behaviour at the top. Something like:

> Certified addresses (CA) are generally more secure than uncertified addresses (UA). Consequently, CAs beat and displace UAs. When the peerstore learns CAs for a peer, it will reject UAs for the same peer (as long as the former haven't expired). Furthermore, peer records act like sequenced snapshots of CAs. Therefore, processing a peer record that's newer than the last one seen overwrites all addresses with the incoming ones.

peerstore/peerstore.go
```diff
@@ -122,6 +123,66 @@ type AddrBook interface {
...
+	// any non-certified addresses added via AddrBook.AddAddrs or
+	// AddrBook.SetAddrs will be ignored. AddrBook.SetAddrs may still be used
+	// to update the TTL of certified addresses if they have previously been
+	// added via AddCertifiedAddrs.
+	AddCertifiedAddrs(s *record.SignedEnvelope, ttl time.Duration) error
+
+	// SignedPeerRecord returns a SignedEnvelope containing a PeerRecord for the
+	// given peer id, if one exists.
+	// Returns nil if no signed PeerRecord exists for the peer.
+	SignedPeerRecord(p peer.ID) *record.SignedEnvelope
```
raulk: `GetPeerRecord`? Just to drop the assumption that the only thing we'll ever do with these records is sign them.

peerstore/peerstore.go
```diff
@@ -122,6 +123,66 @@ type AddrBook interface {
...
+	// adding certified addresses for a peer will *replace* any
+	// existing non-certified addresses for that peer, and only the certified
+	// addresses will be returned from AddrBook.Addrs.
+	//
+	// Likewise, once certified addresses have been added for a given peer,
+	// any non-certified addresses added via AddrBook.AddAddrs or
+	// AddrBook.SetAddrs will be ignored. AddrBook.SetAddrs may still be used
+	// to update the TTL of certified addresses if they have previously been
+	// added via AddCertifiedAddrs.
+	AddCertifiedAddrs(s *record.SignedEnvelope, ttl time.Duration) error
```
raulk: `Add` might be inaccurate here, since this can delete previous addresses. `ProcessPeerRecord`?

record/envelope.go
yusefnapora: I like this, but if there are cases where we construct an `Envelope` without unmarshalling the payload, the `Record` accessor would need to fallback to unmarsahlling if the cached record doesn't exist. In that case, it seems like it should return `(Record, error)` in case unmarshalling fails.

If we did that and had the accessor serve as a "lazy-loading getter", we could potentially skip unmarshalling the record in `ConsumeEnvelope`, and have the unmarshalling always happen in the first call of the accessor function. We could also add a `TypedRecord(dest Record) error` accessor to use a specific `Record` type.

That would let us validate the signature of envelopes whose payloads we don't know how to parse. I'm not sure why/if we want to do that; the only use case I can think of is storing and forwarding `Envelope`s for third-party peers who may know about different payload types than we do.

record/envelope.go
yusefnapora: hehe, I moved it over based on your comment here, https://github.com/libp2p/go-libp2p-core/pull/73#discussion_r361714748, but I'm happy to move it back.

One thing about having the `Seq` field in the `PeerRecord` that's a bit awkward is that, if we remove the `NewPeerRecord` constructor (which I agree is a good idea), then people constructing a `PeerRecord` struct directly will need to explicitly set the `Seq` number, e.g.:

```go
rec := &PeerRecord{PeerID: myId, Addrs: myAddrs, Seq: generateSeqNo()}
```

Or I suppose we could set it for you when you marshal the record if `Seq == 0`, but it feels wrong for a marshal function to alter the thing it's marshaling.

peer/pb/peer_record.proto
raulk: @yusefnapora you may have missed this comment?

peer/record.go
raulk: @yusefnapora You may have missed this comment?

peer/record.go
yusefnapora: sorry, I missed this comment earlier. I think you're right that it makes more sense as a function, now that we have a `Record` interface. 👍

peer/record.go
```diff
@@ -0,0 +1,221 @@
...
+	// Addrs contains the public addresses of the peer this record pertains to.
+	Addrs []ma.Multiaddr
+
+	// Seq is a monotonically-increasing sequence counter that's used to order
+	// PeerRecords in time. The interval between Seq values is unspecified,
+	// but newer PeerRecords MUST have a greater Seq value than older records
+	// for the same peer.
+	Seq uint64
+}
+
```
raulk: Mind adding an interface check here to assert that `PeerRecord` satisfies `record.Record`?

```go
var _ record.Record = (*PeerRecord)(nil)
```

It'll bark if we ever make a change that breaks the interface.

peer/record.go
```diff
@@ -0,0 +1,221 @@
...
+		PeerId:    idBytes,
+		Addresses: addrsToProtobuf(r.Addrs),
+		Seq:       r.Seq,
+	}
+	return proto.Marshal(&msg)
+}
+
+// Sign wraps the PeerRecord in a routing.Envelope, signed with the given
+// private key. The private key must match the PeerID field of the PeerRecord.
+func (r *PeerRecord) Sign(privKey crypto.PrivKey) (*record.Envelope, error) {
```
raulk: Let's not have a `Sign` method per payload type. Proposal:

Add two new methods to the `Record` interface: `Domain() string` and `Codec() []byte`. The `record.MakeEnvelopeWithRecord` could then turn into the `record.Seal(crypto.PrivKey, record.Record)` method we talked about above, and you can remove the redundant parameters, as they can be obtained from the record instance.

record/envelope.go
```diff
@@ -0,0 +1,302 @@
...
+var ErrEmptyPayloadType = errors.New("payloadType must not be empty")
+var ErrInvalidSignature = errors.New("invalid signature or incorrect domain")
+
+// MakeEnvelope constructs a new Envelope using the given privateKey.
+//
+// The required 'domain' string contextualizes the envelope for a particular purpose,
+// and must be supplied when verifying the signature.
+//
+// The 'PayloadType' field indicates what kind of data is contained and must be non-empty.
+func MakeEnvelope(privateKey crypto.PrivKey, domain string, payloadType []byte, payload []byte) (*Envelope, error) {
```
raulk: Both MakeEnvelope methods can be merged if we always pass in an instance of the `Record` interface (once it's been enhanced with the `Codec()` and `Domain()` methods).

record/record.go
```diff
@@ -0,0 +1,88 @@
...
+//
+//    var HelloRecordPayloadType = []byte("/libp2p/hello-record")
+//
+//    func init() {
+//        record.RegisterPayloadType(HelloRecordPayloadType, &HelloRecord{})
+//    }
+//
+//    type HelloRecord struct { } // etc..
+//
+func RegisterPayloadType(payloadType []byte, prototype Record) {
```
raulk: We can change this to `RegisterType(Record)`, where the `Record` interface is self-descriptive, in the way described above.

peer/record.go
```diff
@@ -0,0 +1,221 @@
...
+func init() {
+	record.RegisterPayloadType(PeerRecordEnvelopePayloadType, &PeerRecord{})
+}
+
+// The domain string used for peer records contained in a Envelope.
+const PeerRecordEnvelopeDomain = "libp2p-peer-record"
+
+// The type hint used to identify peer records in a Envelope.
+// TODO: register multicodec
+var PeerRecordEnvelopePayloadType = []byte("/libp2p/peer-record")
```
raulk: We now have codec `0x0301` allocated for peer records: https://github.com/multiformats/multicodec/pull/157

peerstore/peerstore.go
```diff
@@ -122,6 +121,74 @@ type AddrBook interface {
...
+	// adding certified addresses for a peer will *replace* any
+	// existing non-certified addresses for that peer, and only the certified
+	// addresses will be returned from AddrBook.Addrs thereafter.
+	//
+	// Likewise, once certified addresses have been added for a given peer,
+	// any non-certified addresses added via AddrBook.AddAddrs or
+	// AddrBook.SetAddrs will be ignored. AddrBook.SetAddrs may still be used
+	// to update the TTL of certified addresses that have previously been
+	// added via ProcessPeerRecord.
+	ProcessPeerRecord(s *record.Envelope, ttl time.Duration) error
```
raulk: `ConsumePeerRecord`? `Process` sounds kind of generic (I know I had proposed that name, but I'm having second thoughts)

raulk: Can we make this method return (accepted bool, err error), where accepted indicates if this record was accepted?

record/pb/envelope.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package record.pb;
+
+import "crypto/pb/crypto.proto";
+
+// TODO(yusef): doc comments before merge
```
raulk: Now is the time 😆

yusefnapora: haha! good call, will do 😛

peer/record.go
```diff
@@ -0,0 +1,208 @@
...
+// Alternatively, you can construct a PeerRecord struct directly and use the TimestampSeq
+// helper to set the Seq field:
+//
+//    rec := peer.PeerRecord{PeerID: aPeerID, Addrs: someAddrs, Seq: peer.TimestampSeq()}
+//
+// Failing to set the Seq field will not result in an error, however, a PeerRecord with a
+// Seq value of zero may be ignored or rejected by other peers.
+//
+// PeerRecords are intended to be shared with other peers inside a signed
+// routing.Envelope, and PeerRecord implements the routing.Record interface
```
aarshkshah1992: record.Envelope*

record/envelope.go
```diff
@@ -0,0 +1,297 @@
...
+type Envelope struct {
+	// The public key that can be used to verify the signature and derive the peer id of the signer.
+	PublicKey crypto.PubKey
+
+	// A binary identifier that indicates what kind of data is contained in the payload.
+	// TODO(yusef): enforce multicodec prefix
+	PayloadType []byte
+
+	// The envelope payload.
+	RawPayload []byte
```
aarshkshah1992: @yusefnapora Why do we use the `RawPayload ` here instead of using a field of `Record` type since any payload that goes into the Envelope will always be a `Record.`

event/addrs.go
```diff
@@ -0,0 +1,82 @@
...
+	// you whether an address was Added, or was Maintained from the previous
+	// state.
+	Current []UpdatedAddress
+
+	// Removed contains addresses that were removed from the Host.
+	// This field is only set when Diffs == true.
+	Removed []UpdatedAddress
+}
+
+// EvtLocalPeerRoutingStateUpdated should be emitted when a new signed PeerRecord
```
aarshkshah1992: EvtLocalPeerRecordUpdated*

PR #74, opened 2019-11-12
use varints for delimiting plaintext 2.0 msgs

yusefnapora:
This fixes a bug in plaintext 2.0, which according to the spec should use varints for delimiting messages. I used `go-msgio` thinking that it used varints under the hood, but it actually uses 4 byte unsigned ints instead.

This switches out msgio for gogoprotobuf's varint delimited Reader / Writer types.

### DIFF
--- a/sec/insecure/insecure.go
+++ b/sec/insecure/insecure.go
@@ -183,7 +183,7 @@ func (ic *Conn) runHandshakeSync() error {
 
 // read and write a message at the same time.
 func readWriteMsg(rw io.ReadWriter, out *pb.Exchange) (*pb.Exchange, error) {
-	const maxMsgSize = 1 << 32
+	const maxMsgSize = 1 << 16
 	r := gogoio.NewDelimitedReader(rw, maxMsgSize)
 	w := gogoio.NewDelimitedWriter(rw)
 	wresult := make(chan error)
@@ -194,7 +194,7 @@ func readWriteMsg(rw io.ReadWriter, out *pb.Exchange) (*pb.Exchange, error) {
 	inMsg := pb.Exchange{}
 	err := r.ReadMsg(&inMsg)
 
-	// Always wait for the read to finish.
+	// Always wait for the write to finish.
 	err2 := <-wresult
 
 	if err != nil {
--- a/sec/insecure/insecure.go
+++ b/sec/insecure/insecure.go
@@ -6,11 +6,13 @@ package insecure
 import (
 	"context"
 	"fmt"
+	"io"
 	"net"
+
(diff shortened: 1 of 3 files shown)


sec/insecure/insecure.go
```diff
@@ -181,31 +182,28 @@ func (ic *Conn) runHandshakeSync() error {
 }
 
 // read and write a message at the same time.
-func readWriteMsg(c msgio.ReadWriter, out *pb.Exchange) (*pb.Exchange, error) {
-	outBytes, err := out.Marshal()
-	if err != nil {
-		return nil, err
-	}
+func readWriteMsg(rw io.ReadWriter, out *pb.Exchange) (*pb.Exchange, error) {
+	const maxMsgSize = 1 << 32
```
raulk: We can certainly reduce this to e.g. 1024kb, since we know the incoming message should be smaller than that.

sec/insecure/insecure.go
```diff
@@ -181,31 +182,28 @@ func (ic *Conn) runHandshakeSync() error {
...
 	go func() {
-		wresult <- c.WriteMsg(outBytes)
+		wresult <- w.WriteMsg(out)
 	}()
 
-	msg, err1 := c.ReadMsg()
+	inMsg := pb.Exchange{}
+	err := r.ReadMsg(&inMsg)
 
 	// Always wait for the read to finish.
```
raulk:
```suggestion
	// Always wait for the write to finish.
```

sec/insecure/insecure.go
yusefnapora: heh, good call :)

mxinden:
Thanks for fixing this so quickly!

sec/insecure/insecure.go
```diff
@@ -181,31 +182,28 @@ func (ic *Conn) runHandshakeSync() error {
...
 // read and write a message at the same time.
-func readWriteMsg(c msgio.ReadWriter, out *pb.Exchange) (*pb.Exchange, error) {
-	outBytes, err := out.Marshal()
-	if err != nil {
-		return nil, err
-	}
+func readWriteMsg(rw io.ReadWriter, out *pb.Exchange) (*pb.Exchange, error) {
+	const maxMsgSize = 1 << 16
+	r := gogoio.NewDelimitedReader(rw, maxMsgSize)
+	w := gogoio.NewDelimitedWriter(rw)
```
Stebalien, line 188: This will over-read and corrupt the stream.

Stebalien: (it buffers internally)

PR #76, opened 2019-11-21
Fixing go-libp2p#745.

Jorropo:
I'm doing the work for multistream and yamux.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -50,9 +50,17 @@ type Host interface {
 	// (Threadsafe)
 	SetStreamHandler(pid protocol.ID, handler network.StreamHandler)
 
+	// Same as `SetStreamHandler` but return an error if there was already an
+	// handler.
+	SetStreamHandlerSafe(pid protocol.ID, handler network.StreamHandler) error
+
 	// SetStreamHandlerMatch sets the protocol handler on the Host's Mux
 	// using a matching function for protocol selection.
-	SetStreamHandlerMatch(protocol.ID, func(string) bool, network.StreamHandler)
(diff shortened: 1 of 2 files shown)


Jorropo:
@vyzo from libp2p/go-libp2p#745 :
> The problem is if 2 listen are made the second one overide the first (stalling the first, (and a close of the first would stall the second)) but I want listen to return an error if he there is already an stream handler.
So I want AddHandler (or a new func AddHandlerSafe avoiding a refactor) to return error if there is already an stream set (and not setting the new handler).

This just allow not registering a stream if this is is already registered by an other handler.

vyzo:
You are probably doing something wrong in the application if you are registering two competing stream handlers!
Can't you just coordinate that in the application layer? Using a `sync.Once` should resolve this quite easily.

Jorropo:
I know this dumb but this is an user friendly protection, when you try to listen with mutliple process on same port the kernel don't do it and warn you, this is the same here.
Also the current behavior has a major flaw, it is not the object that has a bug that is punished (and its a horror to debug).

And also I hate programs that take initiatives, I don't expect a lib to act like evrything is working when I do something dumb even if for that it have to kill something very probably not expecting that.

Stebalien:
So, I agree that `SetStreamHandler` should reject new handlers for existing protocols. However, this change is not sufficient. Unfortunately, given that handlers registered with `SetStreamHandlerMatch` can handle _arbitrary_ streams, there's no way to _accurately_ detect conflicts.

Given that, I'd like to defer this until we've spent some time to review our interfaces in their entirety.

(known list of warts in my view, there are definitely more: https://github.com/libp2p/go-libp2p/issues/364)

PR #94, opened 2019-12-07
use standard library "crypto/ed25519"

Stebalien:
requires go 1.13


raulk:
Let’s bump the `go` directive in `go.mod` before merging @stebalien.

PR #97, opened 2019-12-11
Update peer

rongyi:
1. fix some staticcheck
2. Drop map default value in AddrInfosFromP2pAddrs
    Go's default behavior is to return the "zero value" for the key not existed in map. there's no need to explicit to do this.

### DIFF
--- a/peer/addrinfo.go
+++ b/peer/addrinfo.go
@@ -29,11 +29,7 @@ func AddrInfosFromP2pAddrs(maddrs ...ma.Multiaddr) ([]AddrInfo, error) {
 		if id == "" {
 			return nil, ErrInvalidAddr
 		}
-		if transport == nil {
-			if _, ok := m[id]; !ok {
-				m[id] = nil
-			}
-		} else {
+		if transport != nil {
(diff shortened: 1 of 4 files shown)


peer/addrinfo.go
```diff
@@ -29,11 +29,7 @@ func AddrInfosFromP2pAddrs(maddrs ...ma.Multiaddr) ([]AddrInfo, error) {
 		if id == "" {
 			return nil, ErrInvalidAddr
 		}
-		if transport == nil {
-			if _, ok := m[id]; !ok {
-				m[id] = nil
-			}
```
Stebalien: This was necessary. Just below this, we iterate over `m` so `id` needs to be a key in the map.

Stebalien: Would you consider adding a test for this?

rongyi: Sorry, you are right, when I changed the code and run the testcase: TestAddrInfosFromP2pAddrs, it passed. But the case does not show this problem and I do find this PR is wrong, I'll close it.

PR #99, opened 2019-12-11
sketch interface for connection gating.

raulk:
For https://github.com/libp2p/go-libp2p/issues/872.

Some downstreams (e.g. Prysmatic, @prestonvanloon, @nisdas) need to actively and strictly limit/deny outbound and inbound connection attempts, after a limit has been reached. On the other hand, other downstreams like Filecoin require peer blacklisting capability.

Enhancing go-libp2p-swarm with peer/connection limit or blacklisting functionality is a dirty solution. The right place to put this concern is in the connection management domain. But not all connection manager implementations will want to act as firewalls or blacklisters.

This PR sketches out a `ConnectionGater` interface that connmgr implementations can implement. The swarm tests if the interface is implemented, and if so, calls `AllowConnection()` before performing any dials (outgoing connection), or incorporating a candidate connection to the swarm (incoming connection). Ideally we'd stop incoming connections at the transport layer, but injecting the connmgr so deep is not something I'm comfortable with.

**TODO:**

- [ ] Gracefully disconnecting the other party -- DISCONNECT protocol: https://github.com/libp2p/go-libp2p/issues/238. If we don't implement this, the other party has no way of knowing if the disconnection was product of a network hiccup, or if it was intentional. And they may try to reconnect.
- [ ] Add the call sites in the swarm.

Before I finish up the remaining work, I'd love to gather some feedback.

Relates to https://github.com/libp2p/go-libp2p-swarm/pull/146.

### DIFF
--- a/connmgr/connmgr.go
+++ b/connmgr/connmgr.go
@@ -65,6 +65,30 @@ type ConnManager interface {
 	Close() error
 }
 
+// ConnectionGater can be implemented by connection managers that support active
+// inbound or outbound connection gating.
+//
+// By default, connection managers are assumed to act like passive sentinels.
+// They sit in the background, watching connections that are established and
+// tracking their score. They perform batch pruning every once in a while to
+// bring the system into equilibrium.
+//
+// A ConnManager implementing ConnectionGater can be consulted by the
+// network.Network (a) before initiating a connection attempt for outbound
+// connections, or (b) incorporating an inbound connection into the node's
+// state. The bool return value indicates whether the caller should proceed.
+//
+// This feature can be used to implement *strict* connection management
+// behaviours, such as hard limiting of connections once a max count has been
+// reached.
+type ConnectionGater interface {
+	// AllowConnection tests whether the given connection is allowed to happen.
+	//
+	// The caller must supply the direction and the peer ID to be tested, and
+	// optionally a network.Conn if one has already been created.
+	AllowConnection(network.Direction, peer.ID, network.Conn) bool
+}
+
 // TagInfo stores metadata associated with a peer.
 type TagInfo struct {
 	FirstSeen time.Time
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
 language: go
 
 go:
-  - 1.11.x
+  - 1.13.x
 
 env:
   global:


raulk:
Lotus needs blacklisting/whitelisting functionality by peer ID, which this interface and the swarm checkpoints are suitable for.

cc @whyrusleeping

prestonvanloon:
Maybe this is another feature, but we want to also stop discovery once the threshold has been met. We believe Kkad DHT discovery is still adding outbound connections which are then disconnected once the event has reached the ConnectionManager and/or network notifee

raulk:
BTW – @prestonvanloon, discussed this with some team members and we do want to warn you that hard limiting connections may be disruptive to protocols that are dynamic and rebalancing in nature. And of course, it may increase the exposure to, or impact of, eclipse attacks.

mcdee:
Prysm keeps its own track of protocol-level peers.  These are peers that have successfully completed a protocol-level handshake to confirm suitability for the nodes to converse.  We also have peers that aren't protocol-level peers but we want to keep around because they perform a function at the network level; an obvious example would be a relay node.

When Prysm reaches the maximum number of protocol-level peers it starts to churn, with peers connecting and starting the protocol-level handshake just for us to tell them to go away.

Although there are some things that we should be doing at the protocol level to tell peers to back off, there's no guarantee that peers will follow the advice so we need to be able to close down incoming peer connections as soon as we can.

Similarly, we need to be able to inform the connection manager not to bother dialling out to a peer if we're just going to reject them when it comes to the protocol-level handshake.

What would be very useful would be to be able to have gating calls both before we dial outbound and before we accept inbound connections.  These would be hooks that would receive details of the connection and would return a boolean "go/no-go" decision.  We can then use the protocol-level information we have to decide if we want to complete the network-level connection and move on to the protocol-level connection.

It seems to me this functionality needs to sit somewhere above swarm but below the connection handler, but I'm no libp2p expert.

aarshkshah1992:
Hey @raulk  I completely understand what we are trying to accomplish here. When do you see this interface PR getting merged  ? We can then start working towards the conn manager implementation, swarm call site changes & the DISCONNECT protocol. I'll share my thoughts on the relevant issues creating them if necessary

raulk:
@mcdee 

> Similarly, we need to be able to inform the connection manager not to bother dialling out to a peer if we're just going to reject them when it comes to the protocol-level handshake.
>
> What would be very useful would be to be able to have gating calls both before we dial outbound and before we accept inbound connections. These would be hooks that would receive details of the connection and would return a boolean "go/no-go" decision. We can then use the protocol-level information we have to decide if we want to complete the network-level connection and move on to the protocol-level connection.

That's indeed the intention. Maybe it wasn't clear in the original description, but that's what I was trying to convey here:

> The swarm tests if the interface is implemented, and if so, calls AllowConnection() before performing any dials (outgoing connection), or incorporating a candidate connection to the swarm (incoming connection). Ideally we'd stop incoming connections at the transport layer, but injecting the connmgr so deep is not something I'm comfortable with.

---

Just one question to make sure we're 100% aligned. In your head, when does _network-level_ stop and _protocol-level_ start?

We won't be able to stop a network connection before the "connection upgrade" is performed, i.e. crypto handshake (from which we learn the peer's identity) and muxer negotiation. Is that good enough for you?

mcdee:
> Just one question to make sure we're 100% aligned. In your head, when does _network-level_ stop and _protocol-level_ start?

We can think of network level as meaning libp2p and protocol level as meaning prysm (in this particular case; in general, the application using libp2p).

> We won't be able to stop a network connection before the "connection upgrade" is performed, i.e. crypto handshake (from which we learn the peer's identity) and muxer negotiation. Is that good enough for you?

I think so, yes.  The main requirement regarding connection gating is for us not to receive messages in prysm from peers that we haven't verified (via a handshake message) as being someone we want to communicate with, which I believe this solution covers.

connmgr/connmgr.go
```diff
@@ -65,6 +65,30 @@ type ConnManager interface {
...
+//
+// This feature can be used to implement *strict* connection management
+// behaviours, such as hard limiting of connections once a max count has been
+// reached.
+type ConnectionGater interface {
+	// AllowConnection tests whether the given connection is allowed to happen.
+	//
+	// The caller must supply the direction and the peer ID to be tested, and
+	// optionally a network.Conn if one has already been created.
+	AllowConnection(network.Direction, peer.ID, network.Conn) bool
```
aarshkshah1992, line 89: @raulk 

Is the `network.Conn` here the connection regarding which a decision is being made ? If so, it will ALWAYS be nil for outbound conns & not-nil for inbound conns.

aarshkshah1992: Infact, the implementation can always fetch the connection for a peer if the Swarm/Network has been injected into it. Given that, should we completely remove the `network.Conn` param from here and simplify the API ?

raulk:
Superseded by https://github.com/libp2p/go-libp2p-core/pull/139 (merged). Work is still in progress to implement this epic across the board. Follow https://github.com/libp2p/go-libp2p/issues/872 for an updated status.

PR #104, opened 2019-12-26
connmgr: introduce abstractions and functions for decaying tags.

raulk:
_Implementation in https://github.com/libp2p/go-libp2p-connmgr/pull/61_

---

This PR introduces a **decaying** _trait_ for the connection manager.

* It is optional for connection managers to implement it.
* The `connmgr.SupportsDecay()` function can be used to check if a given implementation supports decay.
* I'd like to make this pattern more prevalent across go-libp2p.
  * Golang is a terrible language for composability, but I suspect this pattern is reasonable and simple enough to adopt across the board when enhancing existing components (e.g. like with the upcoming Peerstore change to support certified records), without breaking the world for existing implementations.

A decaying connection manager can register _decaying tags_. When doing so, the user provides four things:

* name.
* interval at which the decay ticks.
* a decaying function: takes the old value and returns the new value, or whether the tag should be eliminated.
* a bump function: takes the current value of the tag, and an incoming delta, and returns the new value of the tag.

The solution is decidedly modular and relies on pluggable components to put the power in the hands of users, and to prevent us from overfitting for our particular use cases. Yet, to provide a smooth DX, the upcoming go-libp2p-connmgr PR contributes a pile of useful decay and bump functions.

## Usage

Some example use cases.

### Expiring inactive tags after 10 minutes

```go
decay, ok := connmgr.SupportsDecay(mgr)
if !ok {
     return
}

// this tag ticks every minute, and deletes tags that have been inactive for 10 minutes. Every incoming value overwrites the previous value.
tag, err := decay.RegisterDecayingTag("pop", 1*time.Minute, connmgr.ExpireWhenInactive(10*time.Minute), connmgr.Overwrite())
if err != nil {
    return
}

tag.Bump(<peer id>, 100)
time.Sleep(9 * time.Minute)
tag.Bump(<peer id>, 100)	// refreshes for another 10 minutes
```

### Decrease the value of a tag every minute by half

```go
tag, err := decay.RegisterDecayingTag("pop", 1*time.Minute, connmgr.LinearDecay(0.5), connmgr.Overwrite())
if err != nil {
    return
}

tag.Bump(<peer id>, 100)
time.Sleep(1 * time.Minute)
// value will be 50
time.Sleep(1 * time.Minute)
// value will be 25
```


connmgr/connmgr.go
```diff
@@ -14,14 +14,23 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 )
 
-// ConnManager tracks connections to peers, and allows consumers to associate metadata
-// with each peer.
+// SupportsDecay evaluates if the provided ConnManager supports decay, and if
+// so, it returns the Decayer object. Refer to godocs on Decayer for more info.
+func SupportsDecay(mgr ConnManager) (Decayer, bool) {
```
vyzo: Shouldn't we simply have the extra methods and cast/check the interface?
It feels more idiomatic.

raulk: Isn't that what this does? In the implementation we use type embedding, that's why this works.

vyzo: Fair enough, it amounts to the same thing.
Perhaps we could name the method better though, how about just `Decayer() -> (Decayer, bool)`?

raulk: Name collision, though. In other languages you’d do: `connMgr.as(Decayer.class)`, but no generics :-( I considered CastAsDecayer, ToDecayer, before settling for SupportsDecayer.

vyzo: duh.... How about simply `GetDecayer`?

raulk: I don’t think it makes a substantial difference. “Supports” is more correct because the underlying conn mgr may not support decaying.

vyzo: ok, fine by me.

Stebalien:
What if we provided a helper instead of baking this into the connection manager? That is, every service that needs a decaying tag would wrap the connection manager with a decaying tag manager (for that tag):

```go
tag := DecayingTag(myConnmgr, "my-tag", interval, decayFn, bumpFn)
defer tag.Close() // or cancel with a context?

tag.Bump(peerID, value)
```

Having to call `SupportsDecay` is really awkward. Either:

1. It'll always return a connection manager that supports this feature.
2. The service will have to handle the case when it _doesn't_. That's going to suck.

yusefnapora:
@raulk I've been working against this branch locally but had to rebase against master. I'm going to force-push it up. Just wanted to give a heads-up since it's not my branch :)

For context, we're hoping to get  https://github.com/libp2p/go-libp2p-pubsub/issues/306 sorted out by Monday, which depends on the decaying tag functionality.

raulk:
@yusefnapora I've moved the out-of-the-box decay and bump functions here, so you can now use them without depending on the actual implementation.

I also made the tag interval more lenient, and added `Name()` and `Interval()` methods to the decaying tag, so you can fetch its name and the _effective_ interval at runtime.

PR #107, opened 2020-01-22
Introspection types for core

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/775

All the introspection types/protocols required to be in core.


introspection/providers_tree.go
```diff
@@ -0,0 +1,58 @@
...
+	// ConnID represents a connection ID string
+	ConnID string
+	// StreamID represents a connection ID string
+	StreamID string
+)
+
+// Enums.
+type (
+	// StreamListQueryType is an Enum to represent the types of queries that can be made when looking up streams
+	StreamListQueryType int
```
raulk: I think we can generalise this type and name it `QueryOutput`, with two concrete enum values: `QueryOutputList`, `QueryOutputFull`.

introspection/providers_tree.go
```diff
@@ -0,0 +1,58 @@
+package introspection
+
+// ProvidersTree is a struct which provider modules use to register their interest in providing
+// data to be recorded in a time slice of the running system.
+type ProvidersTree struct {
+	Runtime *RuntimeProviders
```
raulk: I think we can vastly simplify this model by passing in an object to express the type of query we want. If you take a step back and look at the wider pattern here, both `Get`/`List` are really resolver functions that return the same type; what varies is the type of the query, and we don't need to splinter query types into dedicated functions.

```go
type ProvidersMap struct {
    Runtime    func()
    Connection func(ConnectionQueryInput) ([]*Conn, error)
    Stream     func(StreamQueryInput) ([]*Stream, error)
}
```

introspection/providers_tree.go
aarshkshah1992: I've come up with `QueryOutputTypeFull ` & `QueryOutputTypeIds ` which determine whether we want to resolve an output object completely or just fetch it's Ids.

introspection/providers_tree.go
aarshkshah1992: This is a really good idea & as I've gradually begun to understand -> also in the spirit of GraphQL resolvers. 

I've made this change. Also, I've removed the Stream resolver for now as for **0.1**, the streams will be resolved by the connection resolver & the WS client never explicitly asks us to resolve Streams. We can always add it back later when we need it. 

Let me know what you think about the refactored code.

aarshkshah1992:
@raulk  The first round of review has been addressed.

PR #111, opened 2020-02-07
introspection: rename package, rm ListenAddr() from interface.

raulk:
* I think the `introspection` package name is more intuitive.
* The `ListenAddrs()` method shouldn't be part of this interface. An introspector is the thing that performs the introspection, which can be injected in an introspection _server_ to serve the data over an endpoint -- the latter is where the listen addresses belong.


raulk:
Gonna merge this into the feature branch, to make go.mod version management easier.

aarshkshah1992:
@raulk LGTM.

PR #113, opened 2020-02-10
Use `PeerRecord` value instead of pointer  in `EvtLocalPeerRecordUpdated`

yusefnapora:
Small update to #73 that changes the `EvtLocalPeerRecordUpdated` event struct so that it embeds a `PeerRecord` value, instead of a pointer to a `PeerRecord`. This prevents record from being modified by receivers.


raulk:
nit: “embed” has a very specific meaning in go. In this case we’re not embedding the struct, we’re just changing the field type to a value instead of a pointer.

PR #116, opened 2020-02-11
chore(license): fix license detection

### LINKED ISSUE #114: Go docs not displaying on Pkg.go.dev because of "license restrictions"
You can see [here](https://pkg.go.dev/github.com/libp2p/go-libp2p-core?tab=doc) that the docs are not being displayed on the new docs site. It looks like having two licenses is not well supported. I'm sending an email to go-discovery-feedback@google.com about it with a link to this issue and hopefully they'll resolve it. Their site says [here](https://pkg.go.dev/license-policy) that they're using licensecheck but I think it's the code that's using lisensecheck that's at fault, not licensecheck itself.

Stebalien:
Adding URLs to the licenses will cause go to detect this project as Apache 2.0.

fix #114

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,4 @@
+Dual-licensed under MIT and ASLv2, by way of the [Permissive License Stack](https://protocol.ai/blog/announcing-the-permissive-license-stack/).
+
+Apache-2.0: https://www.apache.org/licenses/license-2.0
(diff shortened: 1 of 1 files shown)


Stebalien:
(this is how we're fixing this in go-ipfs)

PR #120, opened 2020-02-28
Introduce EvtPeerConnectednessChanged event

aarshkshah1992:
@Stebalien 

For https://github.com/libp2p/go-libp2p/issues/801


Stebalien:
Connectedness is really a peer concept, connections either exist or don't. For that, I'd emit `EvtPeerStateChange` and emit it when we first form a connection and when we drop the last connection.

Really, I'd just _start_ with that event and only emit stream/conn events _later_ when/if we find we need then. 99% of the time, we _actually_ want to know if we've received/lost a peer, not if we've received/lost a connection/stream.

aarshkshah1992:
@Stebalien 

I understand what you mean and it makes sense. However, given that we ONLY ever have one connection between two peers in libp2p2p(Host.Connect returns immediately if there is an existing connection to the remote peer), is there any difference between

a. We have formed our first connection to a peer
b. We have formed a connection to a peer

On a related note, would the `Host.ConnsToPeer(p peer.ID)` call ever return more than one connection ?

Stebalien:
> ONLY ever have one connection between two peers in libp2p2p

Unfortunately, we frequently have multiple. Peers can connect at the same time, peers can see a new connection before an old one dies, etc.

event/network.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "github.com/libp2p/go-libp2p-core/network"
+
+// EvtPeerStateChange should be emitted everytime we form a connection with a peer or drop our last
+// connection with the peer. Essentially, it is emitted in two cases:
+// a) We form a/any connection with a peer.
+// b) We go from having a connection/s with a peer to having no connection with the peer.
+// It contains the connection handle and the new connection state.
+type EvtPeerStateChange struct {
+	Connection network.Conn
```
Stebalien: I'm not actually sure if the connection should be a part of this event. It should probably just be the peer ID as there could be multiple connections (the service can query for those connections if necessary).

event/network.go
```diff
@@ -0,0 +1,13 @@
...
+	NewState   network.Connectedness
```
Stebalien: Eh, I'd just call this State or Connectedness.

event/network.go
```diff
@@ -0,0 +1,13 @@
+package event
+
+import "github.com/libp2p/go-libp2p-core/network"
+
+// EvtPeerStateChange should be emitted everytime we form a connection with a peer or drop our last
+// connection with the peer. Essentially, it is emitted in two cases:
+// a) We form a/any connection with a peer.
+// b) We go from having a connection/s with a peer to having no connection with the peer.
+// It contains the connection handle and the new connection state.
+type EvtPeerStateChange struct {
```
raulk: `State` is too generic here. If we are talking about connection states, then let's make this explicit in the name.

event/network.go
```diff
@@ -0,0 +1,13 @@
+package event
+
+import "github.com/libp2p/go-libp2p-core/network"
+
+// EvtPeerStateChange should be emitted everytime we form a connection with a peer or drop our last
```
raulk:
```suggestion
// EvtPeerStateChange should be emitted every time we form a connection with a peer or drop our last
```

event/network.go
raulk: `EvtPeerConnectednessChange`?

event/network.go
raulk: Agree; and this could prevent the connection from being GC'ed if the event is queued for too long and we disconnect.

event/network.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+// EvtPeerConnectednessChange should be emitted every time we form a connection with a peer or drop our last
+// connection with the peer. Essentially, it is emitted in two cases:
+// a) We form a/any connection with a peer.
+// b) We go from having a connection/s with a peer to having no connection with the peer.
+// It contains the Id of the remote peer and the new connectedness state.
+type EvtPeerConnectednessChange struct {
+	RemotePeerId  peer.ID
```
raulk: nit: To We already know it's a peer event, not a local event, by its name. Also, we know it's an ID from the type. So in line with all other events, we should rename this field to `Peer`.

raulk: And actually, I just realised another self-nit: it should be `EvtPeerConnectednessChanged` (past).

raulk: See https://github.com/libp2p/go-libp2p-core/blob/master/event/doc.go#L8.

PR #126, opened 2020-03-04
Switch to a single routability event

Stebalien:
Notes/Questions:

* [x] Should we call this "reachability" instead of "routability"?
* [ ] I've put the constants inside the network package for lack of a better place. I assume we'll want to use them elsewhere. For example, we may eventually want the host to answer the question "do I think this peer can be reached from the public internet?"
* Eventually, we'll want something more complex than public/private. We can extend the event when that happens.


network/network.go
```diff
@@ -53,6 +53,24 @@ const (
 	CannotConnect
 )
 
+// Routability indicates how reachable a node is.
+type Routability int
```
willscott: worth making this a `byte`?

event/routability.go
```diff
@@ -1,21 +1,13 @@
...
 
-// EvtLocalRoutabilityUnknown is an event struct to be emitted with the local's
-// node routability changes to UNKNOWN (i.e. we were unable to make a
-// determination about our NAT status with enough confidence).
+// EvtLocalRoutability is an event struct to be emitted with the local's node
+// routability changes state.
 //
 // This event is usually emitted by the AutoNAT subsystem.
-type EvtLocalRoutabilityUnknown struct{}
+type EvtLocalRoutability struct {
```
willscott: maybe `EvtLocalRoutabilityChange`?

network/network.go
aarshkshah1992: @willscott What's the advantage of making this a byte ? Saving space ?

willscott: > @willscott What's the advantage of making this a byte ? Saving space ?

Right. Could potentially allow for better packing

aarshkshah1992:
@Stebalien I like "reachability" more than "routability". Makes the concept more readable & understandable.

event/routability.go
raulk: Yes.

network/network.go
raulk: I'd support it.

event/routability.go
raulk: `EvtLocalRoutabilityChanged` is in tune with the nomenclature proposed here: https://github.com/libp2p/go-libp2p-core/blob/master/event/doc.go#L8

(by the time the consumer receives this event, it's in the past).

network/network.go
Stebalien: No real reason not to.

Stebalien: Hm, well, except for consistency with `Connectedness` and most other enums.

aschmahmann:
@Stebalien anything preventing this from being merged?

network/network.go
```diff
@@ -53,6 +53,25 @@ const (
 	CannotConnect
 )
 
+// Reachability indicates how reachable a node is.
+type Reachability int
```
aschmahmann: Just for clarification is the reason we're put the enum in the network package instead of the event package because we're hoping reachability will be more general than just the `EvtLocalReachabilityChanged` event?

Stebalien: Yes. Basically, it's not specific to the event itself.

PR #130, opened 2020-03-07
deps: move tools to a sub-module

Stebalien:
That way, the root module doesn't depend on gocompat and all its dependencies.


Stebalien:
Merging as this change isn't dangerous, it just removes some deps.

PR #132, opened 2020-03-24
Add an event channel for async Kademlia lookup events.

petar:
Added async event channel following example of query channel.


routing/async.go
```diff
@@ -0,0 +1,128 @@
...
+	kbucket "github.com/libp2p/go-libp2p-kbucket"
+)
+
+// AsyncEvent is emitted for every notable event that happens during a DHT async lookup.
+// AsyncEvent supports JSON marshalling because all of its fields do, recursively.
+type AsyncEvent struct {
+	// ID is a unique identifier for the lookup instance
+	ID string
+	// Key is the Kademlia key used as a lookup target.
+	Key kbucket.ID
```
willscott: you probably want to define your types here and not have core depend on kbucket

generally the dependency structure is
go-libp2p -> [subsystem] -> go-libp2p-core

Stebalien:
If we need a new events system _specifically_ for async kademlia queries, it belongs in the DHT. The routing events system defined in this package is supposed to be a "generic" interface for any routing system (pubsub, DHT, DNS, ec.).

petar:
Moved to go-libp2p-kad-dht.

PR #134, opened 2020-03-25
PeerRecord: Add ToProtobuf() to get a PeerRecord's equivalent protobuf struct

vibhavp:
This lets external protobuf structures import the PeerRecord message type and the Go code using it effectively convert between peer.PeerRecord and its protobuf implementation.


vibhavp:
@Stebalien I've added another function, `PeerRecordFromProtobuf` as well.

vibhavp:
Any update on the PR? Are they any changes needed?

vibhavp:
Thanks!

PR #135, opened 2020-03-26
API to get number of emitters for an event on the bus

aarshkshah1992:
For https://github.com/libp2p/go-eventbus/issues/29.


aarshkshah1992:
Not required for now.

PR #139, opened 2020-04-07
add connection gating interfaces and types.

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/872.

We've decided to model the interface as a bunch of gating functions that are applied during the different life cycle states of the connection. The specific gating function that is applied depends on the lifecycle state of the connection.


connmgr/conngating.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// A ConnectionGater can be consulted by the
+// network.Network (a) before initiating a connection attempt for outbound
+// connections, or (b) incorporating an inbound connection into the node's
+// state.
+//
+// This feature can be used to implement *strict* connection management
+// behaviours, such as hard limiting of connections once a max count has been
+// reached.
+type ConnectionGater interface {
```
aarshkshah1992, line 40: @raulk Once we have the Disconnect message in the Control protocol, I'm inclined to add a `Disconnectpb.Reason` field to the return values of the Gaters so that the caller can send a Disconnect message with the appropriate reason if the connection gets gated.

Is the idea stupid ?

connmgr/conngating.go
```diff
@@ -0,0 +1,43 @@
...
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// AddressConnectionGater can be implemented by a type that supports active
+// connection gating based on the multi-address of the peer.
+type AddressConnectionGater interface {
+	// DenyAddrConnection returns true if a connection to/from a peer
+	// should be denied.
+	// The caller must supply the multi-address of the remote peer.
+	DenyAddrConnection(ma.Multiaddr) (deny bool)
```
willscott, line 16: Is it worth providing both `(ma.Multiaddr, peer.ID)` as arguments. Rather than re-parsing the peer out of the address, most dial attempts ought to be in the context of a specific peer already (so it shouldn't be much of an imposition on the caller), and that might make implementing some type of gaters simpler.

raulk:
I'm having doubts if `AllowConnection` should be tested _after_ negotiating the muxer. In that case, it would be swarm that calls it (in the original spirit of #99), and it could return a disconnect reason, as you suggest, and we could send it given that we now have a muxer.

aarshkshah1992:
@raulk 

In addition to having a muxed connection we can send the `Disconnect` message  on, one more reason I'd consider putting `AllowConnection` in the `Swarm` is to gate outbound connections before they've had the opportunity to travel "downwards" from the `Swarm`.

aarshkshah1992:
@raulk 

Reproducing a question based on an offline discussion

The Connection Gater now allows to gate connections at various stages of it's lifecycle i.e. before it's secured, before a multiplexed is layer on top of it and even after it's completely upgraded.  Let's talk about inbound conns here.

This interface is flexible because it allows transports that don't need upgrading i.e.QUIC etc to gate without having to bubble up an inbound connection all the way to swarm.

However, keep in mind that we also want to send "Disconnect" messages with meaningful reason to the remote peer if we decide to gate & close the inbound conn.

So, for transports that do need "upgrading" i.e. TCP and Websockets, should we skip using the Gater before the upgrade is complete so we can open the Disconnect stream even though the interface allows it ? Or, should we prioritise performance and cost saving here and close the conn as soon as we accept it from the socket or before the multiplexer is layered if we know it needs to die ?

connmgr/conngating.go
```diff
@@ -0,0 +1,64 @@
...
+type ConnectionGater interface {
+	// InterceptDial tests whether we're permitted to dial the specified multiaddr.
+	// Insofar filter.Filters is concerned, this would map to its AddrBlock method,
+	// with the inverse condition.
+	// This is to be called by the network/swarm when dialling.
+	InterceptDial(ma.Multiaddr) (allow bool)
+
+	// InterceptPeerDial tests whether we're permitted to Dial the specified peer.
+	// This is to be called by the network/swarm when dialling.
+	InterceptPeerDial(p peer.ID) (allow bool)
```
aarshkshah1992, line 43: @raulk I've added this function to gate outbound dials based on the peerId. Otherwise, users would have to wait till the dial is complete and InterceptSecured is called. 

Also, gives this functionality to transports that do not use an upgrader.

connmgr/conngating.go
```diff
@@ -0,0 +1,71 @@
...
+// All other methods should allow the connection as we can ONLY open control streams
+// for upgraded connections.
+// Note: There's no point in sending disconnect control messages for outbound connections, so we might
+// as well close them as early in the cycle as possible.
+type ConnectionGater interface {
+	// InterceptDial tests whether we're permitted to dial the specified multiaddr.
+	// Insofar filter.Filters is concerned, this would map to its AddrBlock method,
+	// with the inverse condition.
+	// This is to be called by the network/swarm when dialling.
+	InterceptDial(ma.Multiaddr) (allow bool)
```
Stebalien, line 46: This should pass in the target peer ID as well so we can do things like forbid/allow specific transports/ips to specific peers.

connmgr/conngating.go
```diff
@@ -0,0 +1,71 @@
...
+// behaviours, such as hard limiting of connections once a max count has been
+// reached.
+//
+// If you'd like to send a disconnect control message to the remote peer for a gated inbound connection,
+// ONLY `InterceptUpgraded` should reject the connection (with an appropriate disconnect reason).
+// All other methods should allow the connection as we can ONLY open control streams
+// for upgraded connections.
+// Note: There's no point in sending disconnect control messages for outbound connections, so we might
+// as well close them as early in the cycle as possible.
+type ConnectionGater interface {
```
Stebalien, line 41: I'd consider switching from `allow bool` to a Allow, Deny, Pass system. That way, we can compose multiple connection gaters without having to define new interfaces. This would also allow peers to define a "default" policy.

connmgr/conngating.go
```diff
@@ -0,0 +1,71 @@
...
+	// This is intended to be called by the upgrader, after it has negotiated crypto,
+	// and before it negotiates the muxer, or by the directly by the transport,
+	// at the exact same checkpoint.
+	InterceptSecured(network.Direction, peer.ID, network.ConnMultiaddrs) (allow bool)
+
+	// InterceptUpgraded tests whether a fully capable connection is allowed.
+	// At this point, we have a multiplexer, so the middleware can
+	// return a DisconnectReason and the swarm would use the control stream to convey
+	// it to the peer.
+	InterceptUpgraded(transport.CapableConn) (allow bool, reason control.DisconnectReason)
```
Stebalien, line 70: Given that this is the "final check", why not do this in the swarm and use a normal connection? CapableConn has methods like `AcceptStream` that should only be called by the swarm, as far as I know.

connmgr/conngating.go
```diff
@@ -0,0 +1,71 @@
...
+	// directly (e.g. QUIC, Bluetooth), straight after it's accepted a connection
+	// from its socket.
+	InterceptAccept(network.ConnMultiaddrs) (allow bool)
+
+	// InterceptSecured tests whether a given connection, now authenticated,
+	// is allowed.
+	// This is intended to be called by the upgrader, after it has negotiated crypto,
+	// and before it negotiates the muxer, or by the directly by the transport,
+	// at the exact same checkpoint.
+	InterceptSecured(network.Direction, peer.ID, network.ConnMultiaddrs) (allow bool)
```
Stebalien, line 64: uber nit: `InterceptAuthenticated` (or `InterceptIdentified` but that could be confused with the identify protocol). Not really important.

connmgr/conngating.go
aarshkshah1992: @Stebalien 

> "That way, we can compose multiple connection gaters without having to define new interfaces"

Please can you elaborate on this  ? I was thinking of having a `Compose(gater1, gater2) gater` method so users can compose multiple rules together easily but how does a Allow, Deny, Pass system help there ?

connmgr/conngating.go
aarshkshah1992: Take a look at `InterceptPeerDial `.

connmgr/conngating.go
aarshkshah1992: @Stebalien We do this in the Swarm ONLY:
https://github.com/libp2p/go-libp2p-swarm/blob/316a8557fa84f479d2a247ea4aaf1d7c5fc83812/swarm.go#L195

I'll update the docs to reflect this.

connmgr/conngating.go
Stebalien: That will let me intercept the entire dial to a peer, but it doesn't let me decide "for this peer, should I allow this transport/addr/etc." For example, I may want to relay all outbound connections through a single relay in that case, I'd want to:

1. Allow relay multiaddrs through the target relay.
2. Allow _all_ multiaddrs for the target relay.

connmgr/conngating.go
Stebalien: * Allow -> Immediately allow the dial
* Deny -> Immediately deny the dial.
* Pass/Continue -> Let the next gate check it. Really, this is "no decision" or "does not match any rules".

connmgr/conngating.go
Stebalien: Ah got it. In that case, I'd definitely use the normal `network.Conn` type.

connmgr/conngating.go
aarshkshah1992: @Stebalien I agree with you. Having both the peerId and the addrs here will be helpful here. I am making the change.

connmgr/conngating.go
aarshkshah1992: @Stebalien I am not sure we need to add this additional complexity.

This interface allows users to reject connections based on their lifecycle state. We return true if we want to allow a connection based on the passed info and the current lifecycle state of the connection. No gating function call needs to know anything about any other gating function call. I am not sure what functionality `Pass` will give here/if we need it.

As far as "composing" Gaters in concerned, it's easy to do so with the current implementation by providing a `Compose(gater1, gater2) Gater` that returns a `Gater` implementation that only allows a connection if both the input gaters allow the connection for that lifecycle state.

raulk: @Stebalien If I'm interpreting your proposal correctly, it boils down to adding a HARD ACCEPT action.

I think that breaks composibility and creates invisible coupling between composed gaters. 

We'd like for gaters to only have a local view of the world; they do not know what downstream gaters they would be repressing with their HARD ACCEPT decision.

Imagine a scenario of two gaters:

1. First one accepts "befriended peers".
2. Second one knows the connection limit of the local router and stops accepting connections at N-1, to avoid tripping the router.

If (1) hard accepts connections, (2) might not be given the opportunity to evaluate exactly at the moment of truth, and we might end up tripping up the router.

For that composite to work, (2) would need to precede (1), hence we're implicitly creating spooky action at a distance (i.e. invisible coupling).

Instead, (1) should allow, and (2) should deny. The final result being a DENY.

---

FTR, there are two popular ways one could implement a middleware pattern:

- recursively, by injecting the `next` middleware, and making the current callee call `next` if they want to allow, or return without recursing if they want to deny. This relies on the last `next` being the actual, gated logic.
- iteratively, where middleware components return true/false, and let the framework iterate.

We are choosing the second pattern because it's simpler, and the first pattern is useful in combination with richer APIs, to enable behaviours like "skip next".

connmgr/conngating.go
raulk: Not really necessary. On an outbound dial, the multiaddr contains the peer ID.

connmgr/conngating.go
Stebalien: You've convinced me.

This boils down to ordering. If we say that connection gaters are ordered, then accept, deny, pass make sense. If we're saying they're not ordered, then I agree, we should apply all of them and only accept the connection if they all allow it.

My concern was that I wanted to be able to implement firewall rules. However, the better solution is to just implement a "firewall" connection gater.

```go
fw := Table(
  Rule(Matches("/ip4/127.0.0.1/**"), ResultAllow),
  Rule(Matches("/ip4/6.6.6.6/**"), ResultDeny),
  Rule(Matches(...), Delegate(arbitraryGater)),
  Rule(Matches("/ip6/6.6.6.6/**"), ResultDeny),
  Always(ResultDeny),
)
```

connmgr/conngating.go
raulk: Let's keep the name `Secured` for consistency with `SecureTransport`.

connmgr/conngating.go
raulk: Agree. Changed here: https://github.com/libp2p/go-libp2p-core/pull/139/commits/32264f2111d56b982f47bb6623992e3e78710e89.

raulk:
Reviewed and made changes:

* revised all godocs; made them clearer, and rewrote them in a more imperative, less reflexive manner.
* dropped the `conn` prefix from files in the `connmgr` package.
* marked the `DisconnectReason` type/feature as experimental.
* `InterceptUpgraded(transport.CapableConn)` => `InterceptUpgraded(network.Conn)`.
* `Intercept{Peer => }AddrDial(peer.ID, ma.Multiaddr) (allow bool)`.

PR #145, opened 2020-04-25
fix: envelop contains a mutex and can't be copied.

Stebalien:


aarshkshah1992:
Hey @Stebalien 

Sorry I missed this. Please can you explain why we need this ?

PR #147, opened 2020-04-29
feat: add a function to tell if a context subscribes to query events

Stebalien:
This can save us some work if the caller doesn't actually care about these events.


Stebalien:
Well, it's still the simplest approach. These are _supposed_ to be diagnostic events (which we're then relying on in tests...).

willscott:
... (which we're then relying on in dht commands and hence in return status of HTTP api)

PR #148, opened 2020-05-08
Interface to verify if a peer supports a protocol without making allocations

aarshkshah1992:
Motivation: https://github.com/libp2p/go-libp2p-kad-dht/issues/594.

In a lot of cases, we simply want to know if a peer supports any one protocol among a  set of protocols.
In such cases, we currently end up doing:

```
if s, err := as.host.Peerstore().SupportsProtocols(peer, protos...); 
len(s) == 0 && err == nil {...}
```

However, this causes needless allocations as `SupportsProtocols` makes an allocation for each protocol that is supported even if we don't actually need that information.

This PR adds a `SupportsAnyProtocol ` API that allows clients to determine if the peer supports ANY of the given protocols without knowing which of the given protocols it actually supports.

### DIFF
--- a/peerstore/peerstore.go
+++ b/peerstore/peerstore.go
@@ -238,7 +238,8 @@ type ProtoBook interface {
 	// If the returned error is not nil, the result is indeterminate.
 	SupportsProtocols(peer.ID, ...string) ([]string, error)
 
-	// SupportsAnyProtocol returns true if the peer supports ATLEAST ONE of the given protocols and false otherwise.
+	// FirstSupportedProtocol returns the first protocol that the peer supports among the given protocols.
+	// If the peer does not support any of the given protocols, this function will return an empty string and a nil error.
 	// If the returned error is not nil, the result is indeterminate.
-	SupportsAnyProtocol(peer.ID, ...string) (bool, error)
+	FirstSupportedProtocol(peer.ID, ...string) (string, error)
 }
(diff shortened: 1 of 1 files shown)


peerstore/peerstore.go
```diff
@@ -233,5 +233,12 @@ type ProtoBook interface {
...
 	SetProtocols(peer.ID, ...string) error
 	RemoveProtocols(peer.ID, ...string) error
+
+	// SupportsProtocols returns the set of protocols the peer supports from among the given protocols.
+	// If the returned error is not nil, the result is indeterminate.
 	SupportsProtocols(peer.ID, ...string) ([]string, error)
+
+	// SupportsAnyProtocol returns true if the peer supports ATLEAST ONE of the given protocols and false otherwise.
+	// If the returned error is not nil, the result is indeterminate.
+	SupportsAnyProtocol(peer.ID, ...string) (bool, error)
```
Stebalien, line 243: What about `FirstSupportedProtocol(peer.ID, ...string) (string, error)`? This is what you want 99.9% of the time anyways and covers the new stream use-case.

aarshkshah1992: @Stebalien Great idea !

aarshkshah1992:
@Stebalien Have made the interface change here and in the peerstore PR. Please take a look when you can. Thanks !

Stebalien:
Merge and cut a minor release when ready.

PR #150, opened 2020-05-13
implement Stringer for network.{Direction,Connectedness,Reachability}.

raulk:
Pretty non-controversial.


raulk:
@Stebalien makes sense, I'll rejig this. I also felt kinda uncomfortable panicking instead of returning something like: "(unrecognized)"

PR #153, opened 2020-05-20
eventbus: add wildcard subscription type; getter to enumerate known types

raulk:


raulk:
Gonna merge because I need to make progress; feel free to review after the fact.

PR #155, opened 2020-05-21
overhaul stream closure

raulk:
_Carries over from #10._

---

This PR introduces interface changes to overhaul stream closure.

* `Close()` now behaves like `net.Conn#Close()`, closing the stream fully, with no abnormal termination. It flushes the write buffer, fails future reads, interrupts in-progress reads and writes.
* New `StreamCloseError` struct type to convey optional stream closure errors.
* New `CloseWrite(StreamCloseError)` method to close the write side of the stream only (equivalent to the current `Close()` behaviour -- a footgun for many).
* New `CloseRead(StreamCloseError)` method to close the read side of the stream only. Not all muxers support this; for those that don't this is equivalent to draining the read buffer into `io.Discard`.
* New `ResetErr(StreamCloseError)`, close a stream abnormally just like `Reset()`, but supports passing a `StreamCloseError`.

### DIFF
--- a/mux/mux.go
+++ b/mux/mux.go
@@ -19,9 +19,29 @@ type MuxedStream interface {
 	io.Reader
 	io.Writer
 
-	// Close closes the stream for writing. Reading will still work (that
-	// is, the remote side can still write).
+	// Close closes the stream.
+	//
+	// * Any buffered data for writing will be flushed.
+	// * Future reads will fail.
+	// * Any in-progress reads/writes will be interrupted.
+	//
+	// Close may be asynchronous and _does not_ guarantee receipt of the
+	// data.
 	io.Closer
+
+	// CloseWrite closes the stream for writing but leaves it open for
+	// reading.
+	//
+	// CloseWrite does not free the stream, users must still call Close or
+	// Reset.
+	CloseWrite() error
+
+	// CloseRead closes the stream for writing but leaves it open for
+	// reading.
+	//
+	// CloseRead does not free the stream, users must still call Close or
+	// Reset.
+	CloseRead() error
 
 	// Reset closes both ends of the stream. Use this to tell the remote
 	// side to hang up and go away.
--- a/mux/mux.go
+++ b/mux/mux.go
@@ -27,6 +27,9 @@ type MuxedStream interface {
 	//
 	// Close may be asynchronous and _does not_ guarantee receipt of the
 	// data.
+	//
+	// For adherence with Go conventions, the Close method does not accept
+	// a StreamCloseError. For a detailed
 	io.Closer
 
 	// CloseWrite closes the stream for writing but leaves it open for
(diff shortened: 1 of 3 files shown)


mux/mux.go
```diff
@@ -19,14 +19,55 @@ type MuxedStream interface {
...
+	// reading.
+	//
+	// CloseWrite does not free the stream, users must still call Close or
+	// Reset.
+	//
+	// Users may provide an optional StreamCloseError, to convey a detailed
+	// error to the peer, if the underlying stream muxer supports so. To check
+	// if the underlying stream muxer has used the error, call
+	// StreamCloseError.Used() on return.
+	CloseWrite(err *StreamCloseError) error
```
vyzo, line 45: I don't think it's an error condition to call `CloseWrie` -- the error seems redundant.

mux/mux.go
```diff
@@ -19,14 +19,55 @@ type MuxedStream interface {
...
+	// writing.
+	//
+	// CloseRead does not free the stream, users must still call Close or
+	// Reset.
+	//
+	// Users may provide an optional StreamCloseError, to convey a detailed
+	// error to the peer, if the underlying stream muxer supports so. To check
+	// if the underlying stream muxer has used the error, call
+	// StreamCloseError.Used() on return.
+	CloseRead(err *StreamCloseError) error
```
vyzo, line 57: same here.

mux/mux.go
vyzo: This is equivalent to `shutdown(SHUT_WR)`

mux/mux.go
```diff
@@ -19,14 +19,55 @@ type MuxedStream interface {
...
+	// remote side to hang up and go away.
 	Reset() error
 
+	// ResetErr closes both ends of the stream abruptly, allowing the user to
+	// specify an optional StreamCloseError. If nil, the behaviour is equivalent
+	// to Reset().
+	//
+	// To check if the underlying stream muxer has used the error call
+	// StreamCloseError.Used() on return.
+	ResetErr(err *StreamCloseError) error
```
vyzo, line 69: Again here this proliferation of StremErrors.

Stebalien:
I agree with @vyzo about the errors as discussed in #10. At most, they could be useful for debugging because we can't actually rely on the underlying transport supporting them. Although even that is suspect for security reasons. We probably don't want to send arbitrary errors and users _will_ try to send arbitrary errors.

To make this interface as idiomatic as possible, can we at least make the error types generic `error`s? I'd still like to see an actual motivation and use-case for having errors that's not a foot-gun, but following idioms in the standard library would be better than nothing.

mux/mux.go
raulk: Yes, the intention was to allow `nil` as a value here. All errors sent through this mechanism are purely informational, though. Because they're only sent if the underlying muxer supports it, so they can't be relied upon deterministically.

mux/mux.go
raulk: `StreamError` was a _feature_ of this PR. It's intentional, it's not appearing by magic ;-) There's a plain `Reset()` above that's equivalent to `ResetErr(nil)`.

raulk:
@Stebalien I'm fine dropping the error feature entirely. I agree, these errors are fully informational and cannot be relied upon. Even if we provide the means to verify if the error was actually sent across (like I did through `Used() bool`), one cannot rely on this mechanism to create deterministic behaviour.

This was an attempt to reconcile @marten-seemann's argument that we'd be underusing QUIC's abilities if we forego the ability to pass errors on the various kinds of closures. And I kinda agree.

But I guess this is one of the tradeoffs you have to make when building a modular platform that abstracts implementation details.

raulk:
@Stebalien Feel free to reopen #10. We need to enhance the godocs a little, to make them more instructive, but I'd support that (simpler) approach.

PR #157, opened 2020-05-26
marshalbinarywithouterrors

decanus:


Stebalien:
What's the goal here? `MarshalBinary` is just implementing https://golang.org/pkg/encoding/#BinaryMarshaler. If you want to encode to bytes, just use `[]byte(id)`.

decanus:
Sorry about that @Stebalien wasn’t fully thinking when I opened it

PR #165, opened 2020-08-27
Adding more doc about ClosableTransport.

Jorropo:
Closes (partialy) : libp2p/go-libp2p#999.

### DIFF
--- a/transport/transport.go
+++ b/transport/transport.go
@@ -54,6 +54,15 @@ type CapableConn interface {
 // CapableConn, which means that they have been upgraded to support
 // stream multiplexing and connection security (encryption and authentication).
 //
+// If your transport implements `io.Closer` Close will be called when libp2p
+// closes.
+//
+// Close need to be threadsafe with Dial and Listen.
+// You might receive calls to Dial or Listen even once closed, they should
+// error.
+//
+// If Close return an error the transport is still considered as dead.
+//
 // For a conceptual overview, see https://docs.libp2p.io/concepts/transport/
 type Transport interface {
 	// Dial dials a remote peer. It should try to reuse local listener


transport/transport.go
```diff
@@ -81,6 +81,18 @@ type Transport interface {
 	// See the Network interface for an explanation of how this is used.
 	// TODO: Make this a part of the go-multiaddr protocol instead?
 	Proxy() bool
+	
+	// If your transport implements `io.Closer` Close will be called
```
Stebalien, line 85: This should go above the interface description.

transport/transport.go
```diff
@@ -81,6 +81,18 @@ type Transport interface {
 	// See the Network interface for an explanation of how this is used.
 	// TODO: Make this a part of the go-multiaddr protocol instead?
 	Proxy() bool
+	
+	// If your transport implements `io.Closer` Close will be called 
+	// when libp2p doesn't need it anymore.
+	// (libp2p closing, transport unhooked from a node, ...)
```
Stebalien, line 87: Let's just simplify this to `when the libp2p node shuts down". There's no other way this will happen.

transport/transport.go
```diff
@@ -81,6 +81,18 @@ type Transport interface {
...
+	//
+	// Once a Close is called you can safely assume no more Dial or Listen are
+	// gonna be issued on this instance.
```
Stebalien, line 90: I don't think this is actually guaranteed.

transport/transport.go
```diff
@@ -81,6 +81,18 @@ type Transport interface {
...
+	//
+	// If Close return an error the transport is still considered as dead
+	// (ex: transport already closed or transport never started).
+	//
+	// No more than one Close will ever be issued on an instance.
```
Stebalien, line 95: This is implied in the contract. (or we can extend the statement above to be "If your transport implements `io.Closer` Close will be called exactly once when the libp2p node shuts down".

Stebalien:
https://github.com/libp2p/go-libp2p-core/pull/188

PR #166, opened 2020-08-27
add CloseRead/CloseWrite on streams

### LINKED ISSUE #9: Question: Why doesn't Close() close the stream?

Stebalien:
This changes the behavior of `Close` to behave as one would expect: it closes
the stream. The new methods, CloseWrite/CloseRead allow for closing the stream in
a single direction.

Note: This _does not_ implement CancelWrite/CancelRead as our stream muxer
_protocols_ don't support that.

fixes #9
followup from #115, which is a followup from #10.

* Yamux: https://github.com/libp2p/go-yamux/pull/5
* Mplex: https://github.com/libp2p/go-mplex/pull/81
* QUIC: https://github.com/libp2p/go-libp2p-quic-transport/pull/174


vyzo (approved):
At last! It will be quite an endeavor to update our code for this however.

Stebalien:
> At last! It will be quite an endeavor to update our code for this however.

I've already written the patch for yamux and I'm currently writing the patch for mplex. It's actually pretty easy to update code _using_ streams, because pretty much all of it either:

1. Uses `FullClose` and `AwaitEOF`, which have now been removed.
2. Uses `Reset` (no change).
3. Uses `Close` incorrectly (which will now become "correctly").

raulk (approved):

Stebalien:
The test failures here are just warnings about breaking changes. All muxer PRs are ready to review.

PR #171, opened 2020-11-23
Disovered Addresses Event

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/1012.

AutoNAT PR at https://github.com/libp2p/go-libp2p-autonat/pull/98.

As discussed with @Stebalien , sub-systems that discover new dialable addresses will emit this event which will then be consumed by AutoNAT to determine if they are reachable.

### DIFF
--- a/event/addrs.go
+++ b/event/addrs.go
@@ -80,3 +80,27 @@ type EvtLocalAddressesUpdated struct {
 	// wrapped in a record.Envelope and signed by the Host's private key.
 	SignedPeerRecord *record.Envelope
 }
+
+// DiscoveredAddressesSource represents the source of a new set
+// of discovered dialable addresses.
+type DiscoveredAddressesSource int
+
+const (
+	// InterfaceNewListen means that the dialable addresses are interface listen addresses.
+	InterfaceNewListen DiscoveredAddressesSource = iota
+	// ExternallyObserved means the dialable addresses are addresses observed by remote peers on connections.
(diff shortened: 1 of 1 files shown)


event/addrs.go
```diff
@@ -80,3 +80,27 @@ type EvtLocalAddressesUpdated struct {
...
+	ExternallyObserved
+	// UPnP means the dialable addresses are new addresses assigned by UPnP port mappings.
+	UPnP
+	// NetRoute means the dialable addresses are new interface addresses as determined by NetRoute.
+	NetRoute
+)
+
+// EvtDiscoveredAddresses should be emitted when a new dialable address is discovered by a system.
+// This dialable address would then be verified by a system such as AutoNAT.
+type EvtDiscoveredAddresses struct {
```
willscott, line 101: can we have each event provide a single discovered address?
anyone watching for new addresses will already be wanting to aggregate since addresses may arrive at the same time (e.g. startup) form multiple sources. We can probably reduce the second aggregation here for simplicity

aarshkshah1992: @willscott But imagine a scenario wherein the "default route" changes and we want to send the new interface addresses. Are you saying we'd send two events here for each new interface address ?

willscott: I would prefer getting two events, since the default route changing also likely will occur at the same time as different externally discovered and different netroute events.

PR #173, opened 2021-01-27
Event for user's NAT Device Type: Tell user if the node is behind an Easy or Hard NAT

aarshkshah1992:

Emit event to inform user of the NAT Device Type wrt BAT traversal i.e. Easy NAT(supports consistent endpoint translation) or Hard NAT(does NOT support consistent endpoint translation).

### DIFF
--- a/event/nattype.go
+++ b/event/nattype.go
@@ -2,27 +2,6 @@ package event
 
 import "github.com/libp2p/go-libp2p-core/network"
 
-// NATTransportProtocol is the transport protocol for which the NAT Device Type has been determined.
-type NATTransportProtocol int
-
-const (
-	// NATTransportUDP means that the NAT Device Type has been determined for the UDP Protocol.
-	NATTransportUDP NATTransportProtocol = iota
-	// NATTransportTCP means that the NAT Device Type has been determined for the TCP Protocol.
-	NATTransportTCP
-)
-
-func (n NATTransportProtocol) String() string {
-	switch n {
-	case 0:
-		return "UDP"
-	case 1:
-		return "TCP"
-	default:
-		return "unrecognized"
-	}
-}
-
 // EvtNATDeviceTypeChanged is an event struct to be emitted when the type of the NAT device changes for a Transport Protocol.
 //
 // Note: This event is meaningful ONLY if the AutoNAT Reachability is Private.
@@ -30,7 +9,7 @@ func (n NATTransportProtocol) String() string {
 // this event ONLY if the Reachability on the `EvtLocalReachabilityChanged` is Private.
 type EvtNATDeviceTypeChanged struct {
 	// TransportProtocol is the Transport Protocol for which the NAT Device Type has been determined.
-	TransportProtocol NATTransportProtocol
+	TransportProtocol network.NATTransportProtocol
 	// NatDeviceType indicates the type of the NAT Device for the Transport Protocol.
 	// Currently, it can be either a `Cone NAT` or a `Symmetric NAT`. Please see the detailed documentation
 	// on `network.NATDeviceType` enumerations for a better understanding of what these types mean and
--- a/event/nattype.go
+++ b/event/nattype.go
@@ -12,6 +12,17 @@ const (
 	NATTransportTCP
 )
 
+func (n NATTransportProtocol) String() string {
+	switch n {
+	case 0:
+		return "UDP"
+	case 1:
+		return "TCP"
+	default:
+		return "unrecognized"
+	}
+}
+
 // EvtNATDeviceTypeChanged is an event struct to be emitted when the type of the NAT device changes for a Transport Protocol.
 //
 // Note: This event is meaningful ONLY if the AutoNAT Reachability is Private.
--- a/event/nattype.go
+++ b/event/nattype.go
@@ -2,7 +2,18 @@ package event
 
 import "github.com/libp2p/go-libp2p-core/network"
 
+// NATDeviceProtocol is the transport protocol for which the NAT Device Type has been determined.
+type NATDeviceProtocol int
+
+const (
+	// NATDeviceUDP means that the NAT Device Type has been determined for the UDP Protocol.
+	NATDeviceUDP NATDeviceProtocol = iota
+	// NATDeviceTCP means that the NAT Device Type has been determined for the TCP Protocol.
+	NATDeviceTCP
+)
+
 // EvtNATDeviceTypeChanged is an event struct to be emitted when the type of the NAT device changes.
 type EvtNATDeviceTypeChanged struct {
+	Protocol      NATDeviceProtocol
 	NatDeviceType network.NATDeviceType
 }
--- a/event/nattype.go
+++ b/event/nattype.go
@@ -1,0 +1,8 @@
+package event
+
+import "github.com/libp2p/go-libp2p-core/network"
+
+// EvtNATDeviceTypeChanged is an event struct to be emitted when the type of the NAT device changes.
+type EvtNATDeviceTypeChanged struct {
+	NatDeviceType network.NATDeviceType
+}
--- a/event/nattype.go
+++ b/event/nattype.go
@@ -2,18 +2,27 @@ package event
 
 import "github.com/libp2p/go-libp2p-core/network"
 
-// NATDeviceProtocol is the transport protocol for which the NAT Device Type has been determined.
-type NATDeviceProtocol int
+// NATTransportProtocol is the transport protocol for which the NAT Device Type has been determined.
+type NATTransportProtocol int
 
 const (
-	// NATDeviceUDP means that the NAT Device Type has been determined for the UDP Protocol.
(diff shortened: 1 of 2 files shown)


network/nattype.go
```diff
@@ -0,0 +1,26 @@
+package network
+
+// NATDeviceType indicates the type of the NAT device i.e. whether it is a Hard or an Easy NAT.
+type NATDeviceType int
+
+const (
+	// NATDeviceTypeUnknown indicates that the type of the NAT device is unknown.
+	NATDeviceTypeUnknown NATDeviceType = iota
+
+	// NATDeviceTypeEasy indicates that the NAT device is an Easy NAT i.e. it supports consistent endpoint translation.
```
willscott, line 10: Can we provide the precise definitions of
(a) the heuristic used to determine this
and
(b) what we mean by 'consistent endpoint translation' - do you mean that the externally bound host:port is the same for all remote hosts, stable for different ports of a single remote host, or once mapped allows inbound from a specifically negotiated remote host?

network/nattype.go
```diff
@@ -0,0 +1,26 @@
+package network
+
+// NATDeviceType indicates the type of the NAT device i.e. whether it is a Hard or an Easy NAT.
+type NATDeviceType int
```
willscott, line 4: Is it possible to have this definition of nat type be merged with the existing [network.Reachability](https://github.com/libp2p/go-libp2p-core/blob/3267df19099a1ba68aa32103b13e925efeb2ef04/network/network.go#L74)

this seems to be a sub-division of what's currently used for network.ReachabilityPrivate.

event/nattype.go
```diff
@@ -0,0 +1,19 @@
...
+
+// NATDeviceProtocol is the transport protocol for which the NAT Device Type has been determined.
+type NATDeviceProtocol int
+
+const (
+	// NATDeviceUDP means that the NAT Device Type has been determined for the UDP Protocol.
+	NATDeviceUDP NATDeviceProtocol = iota
+	// NATDeviceTCP means that the NAT Device Type has been determined for the TCP Protocol.
+	NATDeviceTCP
+)
```
marten-seemann, line 13: Why is this in package `event`, but `NATDeviceType` is in `network`?

network/nattype.go
```diff
@@ -0,0 +1,26 @@
...
+	NATDeviceTypeEasy
+
+	// NATDeviceTypeHard indicates that the NAT device is a Hard NAT that does NOT support
+	// consistent endpoint translation.
+	// NAT traversal via hole-punching is NOT possible with this NAT type irrespective of the remote peer's NAT type.
+	NATDeviceTypeHard
+)
+
+func (r NATDeviceType) String() string {
+	str := [...]string{"Unknown", "Easy", "Hard"}
```
marten-seemann, line 21: This allocates, doesn't it? A `switch` statement would be free in terms of memory.

marten-seemann (commented):
Does "consistent endpoint translation" mean the following: For connection originating from IP:port `A`, the NAT will always use `B` as the source 2-tuple on outgoing packets?
That would also mean that if you dial form `A'`, you'd get a different address `B'`, i.e. users have to use SO_REUSEPORT / SO_REUSEADDR in TCP / multiplex connections in QUIC, right?

marten-seemann:
Are "Hard NAT" and "Easy NAT" the terms used in the scientific literature regarding NATs?

aarshkshah1992:
@marten-seemann I've updated the PR to use the NAT terms used in scientific literature and have made the documentation juicier to explain what those NATs mean and how they relate to hole punching. Please let me know what you think.

network/nattype.go
aarshkshah1992: @willscott 

Considering that we need this event on a per transport protocol basis, it'll be extremely tricky to combine this with the `Reachability` event. I've updated the docs to inform the users that they should ALSO consume the Reachability event and use this NAT type event ONLY if Reachability=Private.

network/nattype.go
aarshkshah1992: I've updated the docs to use the terms used in scientific literature and explain consistent endpoint translation in relation to those terms.

network/nattype.go
```diff
@@ -0,0 +1,31 @@
...
+type NATDeviceType int
+
+const (
+	// NATDeviceTypeUnknown indicates that the type of the NAT device is unknown.
+	NATDeviceTypeUnknown NATDeviceType = iota
+
+	// NATDeviceTypeCone indicates that the NAT device is a Cone NAT.
+	// A Cone NAT is a NAT where all outgoing connections from the same source IP address and port are mapped by the NAT device
+	// to the same IP address and port irrespective of the destination address.
+	// With Regards to Internet Society RFC 3489, this could be either a Full Cone NAT, a Restricted Cone NAT or a
```
marten-seemann, line 13:
```suggestion
	// With regards to RFC 3489, this could be either a Full Cone NAT, a Restricted Cone NAT or a
```

network/nattype.go
```diff
@@ -0,0 +1,31 @@
...
+	// Port Restricted Cone NAT. However, we do NOT differentiate between them here and simply classify all such NATs as a Cone NAT.
+	// NAT traversal with hole punching is possible with a Cone NAT if the remote peer is ALSO behind a Cone NAT.
```
marten-seemann, line 15:
```suggestion
	// NAT traversal with hole punching is possible with a Cone NAT, even if the remote peer is also behind a Cone NAT.
```

network/nattype.go
```diff
@@ -0,0 +1,31 @@
...
+	NATDeviceTypeCone
+
+	// NATDeviceTypeHard indicates that the NAT device is a Symmetric NAT.
```
marten-seemann, line 18:
```suggestion
	// NATDeviceTypeSymmetric indicates that the NAT device is a Symmetric NAT.
```

network/nattype.go
```diff
@@ -0,0 +1,31 @@
...
+	// A Symmetric NAT maps outgoing connections with different destination addresses to different IP addresses and ports
```
marten-seemann, line 19:
```suggestion
	// A Symmetric NAT maps outgoing connections with different destination addresses to different IP addresses and ports,
```

marten-seemann (commented):
Documentation is A LOT better now!

network/nattype.go
aarshkshah1992: @marten-seemann 

I've updated the comment to reflect that hole punching will work ONLY if the remote peer is also behind a Cone NAT and fail if the remote peer is behind a Symmetric NAT.

network/nattype.go
aarshkshah1992: Done.

event/nattype.go
aarshkshah1992: Because `event. NATDeviceType ` dosen't make sense. We might have references to just the NATDeviceType in the code elsewhere and storing it as `network.NATDeviceType` is more readable.

We do something similar with `network.Reachability`.

aarshkshah1992:
@marten-seemann 

I've addressed all your comments. Thanks !

event/nattype.go
marten-seemann: That makes sense. But what about `network.NATTransportProtocol`?
Not sure about our package structure here tbh, just asking.

aarshkshah1992: @marten-seemann 

On further thought, I agree with you. Have fixed this by moving the Transport Protocol to the `network` package as well.

aarshkshah1992:
Hey @marten-seemann 


Please approve if happy !

aarshkshah1992:
cc @vyzo for review.

event/nattype.go
```diff
@@ -0,0 +1,18 @@
+package event
+
+import "github.com/libp2p/go-libp2p-core/network"
+
+// EvtNATDeviceTypeChanged is an event struct to be emitted when the type of the NAT device changes for a Transport Protocol.
+//
+// Note: This event is meaningful ONLY if the AutoNAT Reachability is Private.
+// Consumers of this event should ALSO consume the `EvtLocalReachabilityChanged` event and interpret
+// this event ONLY if the Reachability on the `EvtLocalReachabilityChanged` is Private.
+type EvtNATDeviceTypeChanged struct {
```
vyzo, line 10: why `Changed`? It could be the first and only event when it is first detected, there is no change associated with it.

vyzo (commented):
LGTM, just a small objection on the name of the event.

event/nattype.go
aarshkshah1992: Well, I mean, what if the user's WIFI changes or user connects to a new network ?

aarshkshah1992: cc @vyzo  ?

Does this sound reasonable ?

vyzo: I guess we started with `Unknown` so it does change when we detect the actual type, so *FINE*.

aarshkshah1992:
@vyzo Please can you approve the PR ?

vyzo (approved):

PR #177, opened 2021-02-12
fix JSON deserialization of peer.AddrInfo

marten-seemann:


marten-seemann:
Can we turn off / reconfigure Codecov? It's ridiculous that it's failing here.

jacobheun:
> It's ridiculous that it's failing here.

Why is it ridiculous? This PR fixes an issue with deserialization that wasn't caught because it wasn't tested. There's no test case verifying this fix, which makes it entirely possible for this to be an issue in the future if it's left untested.

Codecov configuration may be off, but I think it's a completely valid failure here.

marten-seemann:
> Why is it ridiculous? This PR fixes an issue with deserialization that wasn't caught because it wasn't tested. 

There was no test case for the deserialization in the first place, and there's no test case now, so this PR doesn't make anything worse. It's just modifying untested code.
I have to admit that I don't buy into this whole concept of running a code coverage analyzer as anything more than an informational tool, but that's a different discussion.

jacobheun:
My short response here is that a test should be added verifying this fix does what it says it does. It's low effort, low impact on CI times and prevents regressions of this bug.

Regardless of the existence of code coverage we should be validating patches with tests.

marten-seemann:
Fair enough, I added a test case testing basic encoding / decoding of `AddrInfo`.

peer/addrinfo_serde.go
```diff
@@ -19,19 +19,23 @@ func (pi AddrInfo) MarshalJSON() ([]byte, error) {
...
 	var data map[string]interface{}
-	err := json.Unmarshal(b, &data)
-	if err != nil {
+	if err := json.Unmarshal(b, &data); err != nil {
 		return err
 	}
-	pid, err := IDB58Decode(data["ID"].(string))
-	if err != nil {
-		return err
+	if id, ok := data["ID"]; ok {
```
jacobheun, line 25: Shouldn't this error if there's no ID? I'd expect ```addrInfo.UnmarshalJSON([]byte(`{"Addrs":[]}`))``` or ```addrInfo.UnmarshalJSON([]byte(`{"foo":"bar"}`))``` to error instead of returning nil. A test validating the bad path expectation would be great. No addresses is fine, but no ID isn't.

marten-seemann: Ok, added. PTAL.

peer/addrinfo_serde.go
```diff
@@ -19,19 +20,25 @@ func (pi AddrInfo) MarshalJSON() ([]byte, error) {
...
 	}
-	pi.ID = pid
-	addrs, ok := data["Addrs"].([]interface{})
-	if ok {
-		for _, a := range addrs {
-			pi.Addrs = append(pi.Addrs, ma.StringCast(a.(string)))
+	if addrsEntry, ok := data["Addrs"]; ok {
+		if addrs, ok := addrsEntry.([]interface{}); ok {
+			for _, a := range addrs {
+				pi.Addrs = append(pi.Addrs, ma.StringCast(a.(string)))
```
jacobheun, line 40: PR looks good, the only outstanding issue I see is that `ma.StringCast` will panic if the multiaddr is malformed or we don't know one of the codecs in the multiaddr. Given this might contain an unknown codec perhaps we should just ignore the particular addr and continue trying to unmarshal the others?

Stebalien:
How about https://github.com/libp2p/go-libp2p-core/pull/178? This lets go do all the error/type checking for us.

marten-seemann:
Closing in favor of #178.

PR #178, opened 2021-02-12
refactor: use a helper type to decode AddrInfo from JSON

Stebalien:
This saves us a bunch of error prone manual decode/encode logic.

### DIFF
--- a/peer/addrinfo_serde.go
+++ b/peer/addrinfo_serde.go
@@ -6,33 +6,38 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+// Helper struct for decoding as we can't unmarshal into an interface (Multiaddr).
+type addrInfoJson struct {
+	ID    ID
+	Addrs []string
+}
+
 func (pi AddrInfo) MarshalJSON() ([]byte, error) {
-	out := make(map[string]interface{})
-	out["ID"] = pi.ID.Pretty()
-	var addrs []string
-	for _, a := range pi.Addrs {
-		addrs = append(addrs, a.String())
+	addrs := make([]string, len(pi.Addrs))
+	for i, addr := range pi.Addrs {
+		addrs[i] = addr.String()
 	}
-	out["Addrs"] = addrs
-	return json.Marshal(out)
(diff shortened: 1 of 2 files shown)


marten-seemann:
This works, my fuzzer (not comitted yet) doesn't find any crashes. It requires extra allocations though, not sure if we should care about that.

Stebalien:
> It requires extra allocations though, not sure if we should care about that

Really? I'd expect _fewer_ allocations as we can avoid a lot of interface indirection.

Regardless, I don't think it will matter much either way.

marten-seemann:
This PR:
```
goos: darwin
goarch: amd64
pkg: github.com/libp2p/go-libp2p-core/peer
BenchmarkAddrInfoJSON-16    	  151012	      7853 ns/op	    2787 B/op	      46 allocs/op
```

#177:
```
goos: darwin
goarch: amd64
pkg: github.com/libp2p/go-libp2p-core/peer
BenchmarkAddrInfoJSON-16    	  135700	      8511 ns/op	    3658 B/op	      65 allocs/op
```

benchmark code: 
```go
func BenchmarkAddrInfoJSON(b *testing.B) {
	for i := 0; i < b.N; i++ {
		ai := AddrInfo{ID: testID, Addrs: []ma.Multiaddr{maddrFull}}
		out, err := ai.MarshalJSON()
		if err != nil {
			panic(err)
		}
		var addrInfo AddrInfo
		if err := addrInfo.UnmarshalJSON(out); err != nil {
			panic(err)
		}
		if addrInfo.ID != testID {
			panic(fmt.Sprintf("expected ID to equal %s, got %s", testID.Pretty(), addrInfo.ID.Pretty()))
		}
		if len(addrInfo.Addrs) != 1 || !addrInfo.Addrs[0].Equal(maddrFull) {
			panic(fmt.Sprintf("expected addrs to match %v, got %v", maddrFull, addrInfo.Addrs))
		}
	}
}
```

marten-seemann:
That's unambiguous. Let's use the PR. @Stebalien can you cherry-pick the test case I wrote for #177?

PR #181, opened 2021-02-14
Context to force a direct dial in the Swarm even if we have a Relayed connection

aarshkshah1992:

### DIFF
--- a/network/context.go
+++ b/network/context.go
@@ -12,8 +12,28 @@ var DialPeerTimeout = 60 * time.Second
 
 type noDialCtxKey struct{}
 type dialPeerTimeoutCtxKey struct{}
+type forceDirectDialCtxKey struct{}
 
 var noDial = noDialCtxKey{}
+var forceDirectDial = forceDirectDialCtxKey{}
+
+// EXPERIMENTAL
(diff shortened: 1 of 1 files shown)


vyzo (approved):

network/context.go
```diff
@@ -12,8 +12,28 @@ var DialPeerTimeout = 60 * time.Second
 
 type noDialCtxKey struct{}
 type dialPeerTimeoutCtxKey struct{}
+type forceDirectDialCtxKey struct{}
```
marten-seemann, line 15: Why is the key a struct, not a `const string`?

marten-seemann: I see that it's consistent with what we've been doing so far, just wondering...

aschmahmann: The idea is you can't screw it up by accident. If you use a string someone else could use the string somewhere else and cause a conflict. For example, two applications could use the key word "forceDialDirect" one meaning to skip looking up the peer in some system like a DHT and the other meaning to refuse to use relays. Using an unexported struct makes this type of error basically impossible.

marten-seemann: That's clever. Thanks for the explanation!

marten-seemann (approved):

PR #184, opened 2021-02-19
add a helper function to go directly from a string to an AddrInfo

whyrusleeping:


peer/addrinfo.go
```diff
@@ -61,6 +62,15 @@ func SplitAddr(m ma.Multiaddr) (transport ma.Multiaddr, id ID) {
 	return transport, id
 }
 
+func AddrInfoFromString(s string) (*AddrInfo, error) {
```
willscott, line 65:
```suggestion
// AddrInfoFromString builds an AddrInfo from the string representation of a Multiaddr
func AddrInfoFromString(s string) (*AddrInfo, error) {
```

willscott (approved):

peer/addrinfo.go
Stebalien, line 65: You sure you don't want `AddrInfosFromStrings`?

Stebalien (approved):

peer/addrinfo.go
whyrusleeping: idk, this seems good enough for now? im happy with whatever

PR #191, opened 2021-03-30
fix staticcheck errors

marten-seemann:
This was not fun...


peer/addrinfo_test.go
```diff
@@ -87,8 +87,7 @@ func TestAddrInfosFromP2pAddrs(t *testing.T) {
 	if len(infos) != 0 {
 		t.Fatal("expected no addrs")
 	}
-	infos, err = AddrInfosFromP2pAddrs(nil)
-	if err == nil {
+	if _, err = AddrInfosFromP2pAddrs(nil); err == nil {
```
Stebalien, line 90: nit: in general, when using the `if =` form, use `if :=`. This if statement has side effects which can be very confusing when skimming code.

Stebalien (approved):
Thanks!

PR #193, opened 2021-04-10
make it possible to change ID Hash

xcshuan:
By adjusting these hard-coded parameters to changeable global variables, the code calling go-libp2p-core can change these parameters to meet the requirements.

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -13,6 +13,10 @@ import (
 	mh "github.com/multiformats/go-multihash"
 )
 
+var DefaultIDHash uint64 = mh.SHA2_256
+var DefaultHashPrefix = "Qm"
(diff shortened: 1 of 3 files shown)


aschmahmann:
The definition of a libp2p PeerID is specified in the specs repo (https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md).

Proposals to change the definition of a peerID should occur at the spec level as they are relevant to all libp2p implementations.

xcshuan:
> The definition of a libp2p PeerID is specified in the specs repo (https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md).
> 
> Proposals to change the definition of a peerID should occur at the spec level as they are relevant to all libp2p implementations.

Ok, thanks, I will close the pr.

PR #198, opened 2021-06-28
Routing: Add metric to peer record proto

gidoBOSSftw5731:
For future isMetered preferences, the plan is to just raise this to something like 50000 to significantly lower the priority. Default will be 100, though clients which do not implement this will have a priority of 0. Adding a bool check could work but would be a bit redundant, it would probably be better to say that 0 = 100 and 1 is the lowest metric (read: highest priority) allowed.


Stebalien:
Let's discuss this in https://discuss.libp2p.io/ first.

The general process is:

1. Discuss on the forums (or in a spec issue, but the forums are better for this).
2. Open a spec PR once there's some amount of consensus (https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md).
3. Finally, make a PR here.

But before we even get there, I suggest prototyping locally to see what seems to work for you.

---

When doing this, consider general usefulness. If this is going to become a "core" feature, it needs to be widely applicable. I can't imagine many applications that would want to put a global "connect to me priority" in their peer routing record and we don't want to fill this record definition with one-off fields. 

One solution to this "generalization" problem is to add an extension point. In this case, it would likely be some form of extension map (string -> bytes), or maybe an extension "object" where the protobuf tags are defined to be multicodecs. This would allow applications to extend peer records with arbitrary fields without requiring everyone to agree on what those fields are.

NOTE: This kind of thing _usually_ lives in a "service" and/or "content" routing record. Unfortunately, we don't yet have real "records" for service/content routing, which is why I suggested stashing the field here.

PR #199, opened 2021-07-02
remove minimum Go version from README

marten-seemann:
It's incorrect, and I'm sure we'll forget updating it every half a year. Better to say nothing.


Stebalien (approved):
Yeah. We should probably update the go-libp2p version as well....

PR #200, opened 2021-07-13
ci: use github-actions for compatibility testing

Stebalien:


marten-seemann (approved):
Nice cleanup!

I'm surprised there isn't a compat GitHub Action, which comments on PRs instead of just failing the CI build (which I assume the `gocompat` command does).

Stebalien:
> I'm surprised there isn't a compat GitHub Action, which comments on PRs instead of just failing the CI build (which I assume the gocompat command does).

Now _that_ I'd love to see. Although here, we probably want to just fail CI.

PR #201, opened 2021-07-16
fix: make timestamps strictly increasing

Stebalien:
On Linux, this is almost always the case. Windows, however, doesn't have nanosecond accuracy.

We make the timestamp sequence numbers strictly increasing by returning the last timestamp + 1 where necessary.

### DIFF
--- a/peer/record_test.go
+++ b/peer/record_test.go
@@ -56,11 +56,11 @@ func TestSignedPeerRecordFromEnvelope(t *testing.T) {
 // This is pretty much guaranteed to pass on Linux no matter how we implement it, but Windows has
 // low clock precision. This makes sure we never get a duplicate.
 func TestTimestampSeq(t *testing.T) {
-	last := uint64(0)
+	var last uint64
 	for i := 0; i < 1000; i++ {
 		next := TimestampSeq()
 		if next <= last {
-			t.Errorf("non-increasing timestampfound: %d <= %d", next, last)
+			t.Errorf("non-increasing timestamp found: %d <= %d", next, last)
 		}
 		last = next
 	}
--- a/peer/record_test.go
+++ b/peer/record_test.go
@@ -52,3 +52,16 @@ func TestSignedPeerRecordFromEnvelope(t *testing.T) {
 		}
 	})
 }
+
+// This is pretty much guaranteed to pass on Linux no matter how we implement it, but Windows has
+// low clock precision. This makes sure we never get a duplicate.
+func TestTimestampSeq(t *testing.T) {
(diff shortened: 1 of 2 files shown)


peer/record_test.go
```diff
@@ -52,3 +52,16 @@ func TestSignedPeerRecordFromEnvelope(t *testing.T) {
...
 }
+
+// This is pretty much guaranteed to pass on Linux no matter how we implement it, but Windows has
+// low clock precision. This makes sure we never get a duplicate.
+func TestTimestampSeq(t *testing.T) {
+	last := uint64(0)
+	for i := 0; i < 1000; i++ {
+		next := TimestampSeq()
+		if next <= last {
+			t.Errorf("non-increasing timestampfound: %d <= %d", next, last)
```
marten-seemann, line 63:
```suggestion
			t.Errorf("non-increasing timestamp found: %d <= %d", next, last)
```

peer/record_test.go
```diff
@@ -52,3 +52,16 @@ func TestSignedPeerRecordFromEnvelope(t *testing.T) {
 		}
 	})
 }
+
+// This is pretty much guaranteed to pass on Linux no matter how we implement it, but Windows has
+// low clock precision. This makes sure we never get a duplicate.
+func TestTimestampSeq(t *testing.T) {
+	last := uint64(0)
```
marten-seemann, line 59:
```suggestion
	var last uint64
```

marten-seemann (commented):
1. I searched all locally checked out code and all of GitHub, and it looks like `TimestampSeq` isn't used anywhere. Would removing it be an option?
2. I find the three atomic operations really hard to reason about. A mutex would be a lot simpler.

Stebalien:
> I searched all locally checked out code and all of GitHub, and it looks like TimestampSeq isn't used anywhere. Would removing it be an option?

It's used by the `NewPeerRecord` constructor (here).

> I find the three atomic operations really hard to reason about. A mutex would be a lot simpler.

Hm. Yeah, that's probably fine. This is definitely _not_ a hotspot.

marten-seemann:
> It's used by the NewPeerRecord constructor (here).

I meant outside of the repo. There's no real reason to export it. But I'm ok with not changing that (in this PR at least).

marten-seemann (approved):
That's a lot easier to read, thanks!

Stebalien:
> I meant outside of the repo. There's no real reason to export it. But I'm ok with not changing that (in this PR at least).

Ah. I see... Yeah, probably. But no energy.

PR #204, opened 2021-07-22
remove deprecated Bytes method from the Key interface

marten-seemann:
This method has been deprecated since May 2019.

### DIFF
--- a/crypto/ecdsa.go
+++ b/crypto/ecdsa.go
@@ -12,7 +12,7 @@ import (
 
 	pb "github.com/libp2p/go-libp2p-core/crypto/pb"
 
-	sha256 "github.com/minio/sha256-simd"
+	"github.com/minio/sha256-simd"
 )
 
 // ECDSAPrivateKey is an implementation of an ECDSA private key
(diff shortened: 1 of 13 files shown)


Stebalien (approved):

Stebalien:
@marten-seemann it looks like the deprecation here wasn't quite "correct". Unfortunately, we have quite a few packages that still use it.

I'm fixing them, but we might want to consider reverting before releasing (for now).

marten-seemann:
> @marten-seemann it looks like the deprecation here wasn't quite "correct". Unfortunately, we have quite a few packages that still use it.

Doesn't that just mean that we were negligent fixing our own code? Why would we revert this, if we can just fix it (replacing `key.Bytes()` by `crypto.Marshal{Private,Public}Key(key)` is not that hard (just tedious).

Stebalien:
> Doesn't that just mean that we were negligent fixing our own code?

We signaled deprecation with "DEPRECATED" instead of "Deprecated: " so `go vet` didn't catch it.

PR #207, opened 2021-07-22
feat: keep addresses for longer

Stebalien:
There's really no reason to expire these after 10 minutes as opposed to 30. Given that our default DHT refresh interval is 10 minutes, setting this to 30 reduces the chances that we'll "forget" peers.

### DIFF
--- a/peerstore/peerstore.go
+++ b/peerstore/peerstore.go
@@ -28,14 +28,14 @@ var (
 	// ProviderAddrTTL is the TTL of an address we've received from a provider.
 	// This is also a temporary address, but lasts longer. After this expires,
 	// the records we return will require an extra lookup.
-	ProviderAddrTTL = time.Minute * 10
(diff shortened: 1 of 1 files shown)


aschmahmann (approved):
This seems like a reasonable idea to me. I know peerstore numbers are a bit magical, so I'm not sure if there's other fallout from here.

cc @marten-seemann as a heads up

aschmahmann:
Note: maybe I'm paranoid, but this might be the kind of PR we want to bubble up + test in larger binaries before merging. On the other hand, it's a pretty easy revert if we need to and this kind of change should be fine.

Stebalien:
I _really_ doubt this will cause significant issues, honestly.

aschmahmann:
SGTM

Stebalien:
Well, it could affect https://github.com/libp2p/go-libp2p/issues/351. But I don't think it'll make anything much _worse_.

PR #208, opened 2021-07-22
feat: remove unused metrics

### LINKED ISSUE #58: Remove opencensus dep
Unfortunately, this means that anything wanting to use the libp2p interfaces (even light-clients that leave all the heavy lifting up to a daemon) need to import opencensus.

From the looks of it, we aren't actually using any of the code in `metrics/register.go`. We might want to consider removing this code unless we think we're actually going to use it.

Note: this pulls in stuff like grpc, google's cloud API libraries, etc.

Stebalien:
This drops the opensensus dependency. Nobody uses this, from what I can tell.

fixes #58


marten-seemann:
What about https://github.com/ipfs/go-ipfs/blob/master/core/corehttp/metrics.go? Note: I don't understand why we use OpenCensus there.

Stebalien:
We were trying to switch over to OpenCensus at one point. I think that's still on the roadmap.

I just want to remove the dependency from the core interfaces. go-ipfs will still depend on it, but users can pull in go-libp2p-core (e.g., because they need some random type) without it.

marten-seemann (approved):

PR #211, opened 2021-09-05
pass the peer ID to SecureInbound in the SecureTransport and SecureMuxer

marten-seemann:
The peer ID may be empty. This will be the common case. In that case, connections from any peer are accepted.

See https://github.com/libp2p/go-libp2p-transport-upgrader/pull/82#discussion_r702406483.


marten-seemann:
Note: I intend to cut a new go-libp2p-core release (v0.10.0) right after this PR is merged. This will allow us to make the necessary changes to -tls, -noise, -upgrader etc.

Stebalien (approved):

PR #212, opened 2021-09-05
remove the Process from the Network interface, add Close method

marten-seemann:


Stebalien (approved):
This really never should have been exposed as there was no "safe" way to use it (there's always a race with shutdown, which can lead to panics). LGTM but you'll probably need to modify quite a few upstream pieces before we can merge (e.g., the DHT, I think?).

marten-seemann:
> LGTM but you'll probably need to modify quite a few upstream pieces before we can merge (e.g., the DHT, I think?).

Yes, this will unfortunately be a larger change. I was thinking of not bundling it in the v0.10 release (which I'll cut after #211), so we can move these changes forward independently.

marten-seemann:
>  I was thinking of not bundling it in the v0.10 release (which I'll cut after #211), so we can move these changes forward independently.

I changed my mind. Using `testing.Cleanup` in https://github.com/libp2p/go-libp2p-swarm/pull/278 made this a lot less invasive.

PR #213, opened 2021-09-06
make Go compat post a comment on the PR

marten-seemann:
`gocompat` currently fails the build when an incompatible API change was made. This is annoying. It's would be a lot more helpful if it would just the author (and the reviewers) what's going on, so they can make an informed decision.

Example:
image

Related: https://github.com/protocol/.github/issues/188


marten-seemann:
Requesting a review from @mvdan (related: https://github.com/protocol/.github/issues/188).

Stebalien (approved):

PR #215, opened 2021-09-21
change interfaces for the security-in-multiaddr change

marten-seemann:
## Changes to the `Transport.Protocol` interface

`Protocols()` now returns a `[]ma.Protocol`, not a slice of untyped ints. Note that this also changes the interpretation if more than one element is returned: So far, returning multiple protocols here meant that the `Transport` supports multiple protocols at the same time. Now it means that the `Transports` supports these protocols **on top of each other**.
Use case: We will have one TCP Transport that handles TLS and one TCP Transport that handles Noise handshakes (the handshake protocol will be taking from the multiaddr, and not negotiated using multistream). These transports will then return `[ TCP, TLS ]` and `[ TCP, Noise ]`, respectively.

## Changes to the `SecureTransport` interface

This PR adds `Protocol` method, returning the handshake protocol this `SecureTransport` uses. Examples: plaintextv2, TLS, Noise.

### DIFF
--- a/sec/insecure/insecure.go
+++ b/sec/insecure/insecure.go
@@ -9,12 +9,13 @@ import (
 	"io"
 	"net"
 
+	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/sec"
-	"github.com/libp2p/go-msgio"
-
-	ci "github.com/libp2p/go-libp2p-core/crypto"
 	pb "github.com/libp2p/go-libp2p-core/sec/insecure/pb"
+
+	"github.com/libp2p/go-msgio"
+	ma "github.com/multiformats/go-multiaddr"
 )
 
 // ID is the multistream-select protocol ID that should be used when identifying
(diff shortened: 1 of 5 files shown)


sec/insecure/insecure.go
```diff
@@ -28,28 +29,36 @@ const ID = "/plaintext/2.0.0"
 // No authentication of the remote identity is performed.
 type Transport struct {
 	id  peer.ID
-	key ci.PrivKey
+	key crypto.PrivKey
```
raulk, line 32: The reason we use a different import alias is to differentiate from the Go SDK crypto. I don't have strong opinions either way, just noting it.

transport/transport.go
```diff
@@ -75,10 +75,10 @@ type Transport interface {
 	// Listen listens on the passed multiaddr.
 	Listen(laddr ma.Multiaddr) (Listener, error)
 
-	// Protocol returns the set of protocols handled by this transport.
-	//
+	// Protocols returns the set of protocols handled by this transport.
+	// If protocols A and B are returned, this means that this transport supports running B on top of A.
```
raulk, line 79: Is this a semantic change, or was the previous interpretation supposed to be stacked too?

raulk (approved):

transport/transport.go
raulk: Oh wait, I just read the PR description. I think we will find this interface limiting, and suggest something more sophisticated. [][]protocol.ID could be an option, but we might want to go with a struct for more developer friendliness.

marten-seemann: That’s correct, it is a semantic change (as I’ve noted in the PR description). None of our transports ever returned more than one value here though.

marten-seemann: Just read your 2nd comment after posting my reply. We could define a `ProtocolStack` as `[]Protocol`, and then return a `[]ProtocolStack` here.

One example where returning multiple protocols might be useful is when we mint a new code point for a new QUIC version. On the other hand, one might argue that if two code points are warranted, the differences between the two transports are probably large enough to just initializing two objects.

raulk: That would be more self-descriptive, yes. Not sure if ProtocolStack is the right term --  since in the past we've talked about libp2p being a factory of _stacks_ with different characteristics and API shapes/semantics (e.g. message orientation). So that might get overloaded in the future. Perhaps ProtocolSeq?

raulk: Also, the term "stack" to me denotes some kind of behaviour associated with that type, versus that term just being a keyword to enable AND and OR combinations of protocols.

PR #219, opened 2021-11-11
generate ecdsa public key from an input public key

richard-ramos:


marten-seemann (commented):
This is not incorrect, but where do we need this? We don't have a corresponding method for RSA, for example.

richard-ramos:
I created this function because currently I'm working on a project where I need to generate a `peer.ID` using a known ecdsa public key, and found that there was no way to initialize a `ECDSAPublicKey` from a `ecdsa.PublicKey`. With this PR I could do something like this:
```go
pubKey, err := ECDSAPublicKeyFromPubKey(someECDSAPubKey)
thePeerId, err := peer.IDFromPublicKey(pubKey)
```
Should I implement a similar function for RSA and Ed25519 too?

marten-seemann:
> I created this function because currently I'm working on a project where I need to generate a peer.ID using a known ecdsa public key, and found that there was no way to initialize a ECDSAPublicKey from a ecdsa.PublicKey.

Ok, that's legit.

> Should I implement a similar function for RSA and Ed25519 too?

No need to.

Can we have a test please?

richard-ramos:
I've added the required test

crypto/ecdsa.go
```diff
@@ -67,6 +67,15 @@ func ECDSAKeyPairFromKey(priv *ecdsa.PrivateKey) (PrivKey, PubKey, error) {
 	return &ECDSAPrivateKey{priv}, &ECDSAPublicKey{&priv.PublicKey}, nil
 }
 
+// ECDSAPublicKeyFromPubKey generates a new ecdsa public key from an input public key
+func ECDSAPublicKeyFromPubKey(pub *ecdsa.PublicKey) (PubKey, error) {
```
aschmahmann, line 71: From some conversation during triage:

It is likely better to dereference they key before it's passed in to reduce the likelihood of hard to track down bugs. Also the keys aren't too big

crypto/ecdsa.go
```diff
@@ -67,6 +67,15 @@ func ECDSAKeyPairFromKey(priv *ecdsa.PrivateKey) (PrivKey, PubKey, error) {
...
+	if pub == nil {
+		return nil, ErrNilPublicKey
+	}
+
+	return &ECDSAPublicKey{pub}, nil
+}
```
Jorropo, line 77: For context, the bug I see here is that someone could pass a key into it and since it's by reference modify it while libp2p is using it (libp2p expects key to be constant and not change).

If it's passed by value then a copy is made and this can't happen.
```suggestion
func ECDSAPublicKeyFromPubKey(pub ecdsa.PublicKey) (PubKey, error) {
  return &ECDSAPublicKey{pub:&pub}, nil
}
```

richard-ramos:
Thank you for the code review, @aschmahmann and @Jorropo . I have done the requested changes.

Should I open a separate PR for doing the same change in here?:
```go
// ECDSAKeyPairFromKey generates a new ecdsa private and public key from an input private key
func ECDSAKeyPairFromKey(priv *ecdsa.PrivateKey) (PrivKey, PubKey, error) {
	if priv == nil {
		return nil, nil, ErrNilPrivateKey
	}

	return &ECDSAPrivateKey{priv}, &ECDSAPublicKey{&priv.PublicKey}, nil
}
```

Jorropo:
@richard-ramos no you should rebase or amend commit.

First, on your `ecdsa-pubkey` create a commit fixing that and do a `git rebase -i master` and read the comments, you will want to `fixup` the new commit second commit into the first one, then we can `git push --force` this branch.

This will change this PR to a new commit that is just like you did it right the first time (sometime you want to do "history" based commits so you would just make a new commit and explain your thought process in the commit description, for something that small it's obvious you can just force push).

richard-ramos:
@Jorropo but i did rebase this PR. Notice it's only a single commit containing the suggested changes

marten-seemann (approved):

PR #222, opened 2021-11-28
move the event.Bus to eventbus.Bus, add a Start method to the Peerstore

marten-seemann:
I'm a bit sad about this change (`event.Bus` just read so nice), but it's causing an import loop if we want to add a `Start(event.Bus)` method to the `Peerstore`.

We need this method to run the garbage collection Go routine in the peerstore (which removes disconnected peers from the peerstore after a grace period), see https://github.com/libp2p/go-libp2p-peerstore/pull/180.

The alternative to this approach would be to pass the `event.Bus` to the constructor of the `Peerstore` implementation. Arguably, this would be the nicer design. That way, we wouldn't need the interface to import the `event` package, avoiding the import cycle.
However, this conflicts with the assumption that the peerstore is a "standalone component" that can be constructed without any other components of the host. For example, this assumption is used go-libp2p to define the `DefaultPeerstore` option: https://github.com/libp2p/go-libp2p/blob/3a71a6d0591482de51b6eaa94093804ea21ab774/defaults.go#L49-L55.

Closes #221.

### DIFF
--- a/event/bus_deprecated.go
+++ b/event/bus_deprecated.go
@@ -1,0 +1,34 @@
+package event
+
+import (
+	"github.com/libp2p/go-libp2p-core/eventbus"
+)
+
+// SubscriptionOpt represents a subscriber option. Use the options exposed by the implementation of choice.
+// Deprecated: Use eventbus.SubscriptionOpt
+type SubscriptionOpt = eventbus.SubscriptionOpt
+
+// EmitterOpt represents an emitter option. Use the options exposed by the implementation of choice.
+// Deprecated: Use eventbus.EmitterOpt
+type EmitterOpt = eventbus.EmitterOpt
+
+// CancelFunc closes a subscriber.
+// Deprecated: Use eventbus.CancelFunc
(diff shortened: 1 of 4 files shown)


Stebalien:
Ok, so the import cycle is:

1. `event` imports `network` for "Reachability".
2. `network` imports `peerstore` for the `Peerstore` method (which I'd kind of like to remove in a future refactor...).
3. `peerstore` now needs to import `event.Bus`.

What if we just moved the definition of `Reachability` somewhere else (e.g., into an internal pacakge) and type-aliased it into `network`? That would let `event` and `network` import this common type.

> However, this conflicts with the assumption that the peerstore is a "standalone component" that can be constructed without any other components of the host.

Well, personally I _would_ make it a constructor option but make it optional. When start is later called, GC would be enabled if the peerstore was constructed with an event bus.

Why? Well... it works better with dependency injection (which I still want to eventually add). Usually, with dependency injection, you'd register a constructor that "provides" a Peerstore (returns it), then register a set of services that optionally depend on the peerstore.

PR #224, opened 2021-12-10
return a ConnectionStat instead of a Stat for the Conn

marten-seemann:
This allows us to expose the number of streams that are currently open on a connection, which will allow us to prioritize killing of connections with many streams in the connection manager.

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -96,7 +96,14 @@ func (r Reachability) String() string {
 	return str[r]
 }
 
-// Stat stores metadata pertaining to a given Stream/Conn.
+// ConnectionStat stores metadata pertaining to a given Conn.
+type ConnectionStat struct {
+	Stat
+	// NumStreams is the number of streams on the connection.
+	NumStreams int
+}
+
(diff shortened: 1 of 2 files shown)


network/network.go
```diff
@@ -96,7 +96,14 @@ func (r Reachability) String() string {
 	return str[r]
 }
 
-// Stat stores metadata pertaining to a given Stream/Conn.
+// ConnectionStat stores metadata pertaining to a given Conn.
+type ConnectionStat struct {
```
vyzo, line 100: Let's call it `ConnStat` for consistency?

vyzo (approved):
We probably should call it `ConnStat` as we don't spell it fully anywhere; other than that LGTM.

network/network.go
marten-seemann: I'd love to, but `ConnStat` is already the interface: https://github.com/libp2p/go-libp2p-core/blob/975f71b238330cbfafb4a96f674ff047686dc219/network/conn.go#L63-L67

vyzo:
Ouch.... oh well.

PR #228, opened 2021-12-15
remove duplicate io.Closer on Network interface

YRXING:
Is there a special purpose to define two `io.Closer` in the `Network` interface?


marten-seemann (approved):
Thank you!

PR #229, opened 2021-12-16
Network Resource Manager interface

vyzo:
See https://github.com/libp2p/go-libp2p/issues/1260 for the proposal/design discussion.

### DIFF
--- a/network/conn.go
+++ b/network/conn.go
@@ -32,7 +32,7 @@ type Conn interface {
 	GetStreams() []Stream
 
 	// Scope returns the user view of this connection's resource scope
-	Scope() UserConnectionScope
+	Scope() ConnectionScope
 }
 
 // ConnSecurity is the interface that one can mix into a connection interface to
--- a/network/conn.go
+++ b/network/conn.go
@@ -30,6 +30,9 @@ type Conn interface {
 
 	// GetStreams returns all open streams over this conn.
 	GetStreams() []Stream
+
+	// Scope returns the connection resource scope
+	Scope() ConnectionScope
 }
 
 // ConnSecurity is the interface that one can mix into a connection interface to
--- a/network/conn.go
+++ b/network/conn.go
@@ -20,6 +20,7 @@ type Conn interface {
 	ConnSecurity
 	ConnMultiaddrs
 	ConnStat
+	ConnScope
 
 	// ID returns an identifier that uniquely identifies this Conn within this
 	// host, during this run. Connection IDs may repeat across restarts.
@@ -30,9 +31,6 @@ type Conn interface {
 
 	// GetStreams returns all open streams over this conn.
 	GetStreams() []Stream
-
-	// Scope returns the user view of this connection's resource scope
-	Scope() ConnectionScope
 }
 
 // ConnSecurity is the interface that one can mix into a connection interface to
@@ -67,4 +65,11 @@ type ConnMultiaddrs interface {
 type ConnStat interface {
 	// Stat stores metadata pertaining to this conn.
 	Stat() ConnStats
+}
+
+// ConnScope is the interface that one can mix into a connection interfac to give it a resource
+// management scope
+type ConnScope interface {
+	// Scope returns the user view of this connection's resource scope
+	Scope() ConnectionScope
 }
--- a/network/conn.go
+++ b/network/conn.go
@@ -67,7 +67,7 @@ type ConnStat interface {
 	Stat() ConnStats
 }
 
-// ConnScoper is the interface that one can mix into a connection interfac to give it a resource
+// ConnScoper is the interface that one can mix into a connection interface to give it a resource
 // management scope
 type ConnScoper interface {
 	// Scope returns the user view of this connection's resource scope
--- a/network/conn.go
+++ b/network/conn.go
@@ -31,8 +31,8 @@ type Conn interface {
 	// GetStreams returns all open streams over this conn.
 	GetStreams() []Stream
 
-	// Scope returns the connection resource scope
-	Scope() ConnectionScope
+	// Scope returns the user view of this connection's resource scope
+	Scope() UserConnectionScope
 }
 
 // ConnSecurity is the interface that one can mix into a connection interface to
--- a/network/conn.go
+++ b/network/conn.go
@@ -20,7 +20,7 @@ type Conn interface {
 	ConnSecurity
 	ConnMultiaddrs
 	ConnStat
-	ConnScope
+	ConnScoper
 
 	// ID returns an identifier that uniquely identifies this Conn within this
 	// host, during this run. Connection IDs may repeat across restarts.
@@ -67,9 +67,9 @@ type ConnStat interface {
 	Stat() ConnStats
 }
 
-// ConnScope is the interface that one can mix into a connection interfac to give it a resource
+// ConnScoper is the interface that one can mix into a connection interfac to give it a resource
 // management scope
-type ConnScope interface {
+type ConnScoper interface {
 	// Scope returns the user view of this connection's resource scope
-	Scope() ConnectionScope
+	Scope() ConnScope
 }
--- a/network/errors.go
+++ b/network/errors.go
@@ -1,6 +1,17 @@
 package network
 
-import "errors"
+import (
+	"errors"
+	"net"
+)
+
+type TemporaryError string
+
+func (e TemporaryError) Error() string   { return string(e) }
+func (e TemporaryError) Temporary() bool { return true }
+func (e TemporaryError) Timeout() bool   { return false }
+
+var _ net.Error = TemporaryError("")
 
 // ErrNoRemoteAddrs is returned when there are no addresses associated with a peer during a dial.
 var ErrNoRemoteAddrs = errors.New("no remote addresses")
@@ -12,3 +23,11 @@ var ErrNoConn = errors.New("no usable connection to peer")
 // ErrTransientConn is returned when attempting to open a stream to a peer with only a transient
 // connection, without specifying the UseTransient option.
 var ErrTransientConn = errors.New("transient connection to peer")
+
+// ErrResourceLimitExceeded is returned when attempting to perform an operation that would
+// exceed system resource limits.
+var ErrResourceLimitExceeded = TemporaryError("resource limit exceeded")
+
+// ErrResourceScopeClosed is returned when attemptig to reserve resources in a closed resource
+// scope.
+var ErrResourceScopeClosed = errors.New("resource scope closed")
--- a/network/errors.go
+++ b/network/errors.go
@@ -5,13 +5,13 @@ import (
 	"net"
 )
 
-type TemporaryError string
+type temporaryError string
 
-func (e TemporaryError) Error() string   { return string(e) }
-func (e TemporaryError) Temporary() bool { return true }
-func (e TemporaryError) Timeout() bool   { return false }
+func (e temporaryError) Error() string   { return string(e) }
+func (e temporaryError) Temporary() bool { return true }
+func (e temporaryError) Timeout() bool   { return false }
 
-var _ net.Error = TemporaryError("")
+var _ net.Error = temporaryError("")
 
 // ErrNoRemoteAddrs is returned when there are no addresses associated with a peer during a dial.
 var ErrNoRemoteAddrs = errors.New("no remote addresses")
@@ -26,7 +26,7 @@ var ErrTransientConn = errors.New("transient connection to peer")
 
 // ErrResourceLimitExceeded is returned when attempting to perform an operation that would
 // exceed system resource limits.
-var ErrResourceLimitExceeded = TemporaryError("resource limit exceeded")
+var ErrResourceLimitExceeded = temporaryError("resource limit exceeded")
 
 // ErrResourceScopeClosed is returned when attemptig to reserve resources in a closed resource
 // scope.
--- a/network/rcmgr.go
+++ b/network/rcmgr.go
@@ -14,12 +14,14 @@ var (
 
 // ResourceManager is the interface to the network resource management subsystem
 type ResourceManager interface {
-	// GetProtocolScope retrieves the resource management scope for a specific protocol.
+	// GetService retrieves a service-specific scope
+	GetService(srv string) ServiceScope
+	// GetProtocol retrieves the resource management scope for a specific protocol.
 	// If there is no configured limits for a particular protocol, then the default scope is
 	// returned.
-	GetProtocolScope(protocol.ID) ProtocolScope
-	// GetPeerScope retrieces the resource management scope for a specific peer.
-	GetPeerScope(peer.ID) PeerScope
+	GetProtocol(protocol.ID) ProtocolScope
+	// GetPeer retrieces the resource management scope for a specific peer.
+	GetPeer(peer.ID) PeerScope
 
 	// OpenConnection creates a connection scope not yet associated with any peer
 	OpenConnection(dir Direction, usefd bool) (ConnectionScope, error)
@@ -37,7 +39,9 @@ type ResourceScope interface {
 
 	// GetBuffer reserves memory and allocates a buffer through the buffer pool.
 	GetBuffer(size int) ([]byte, error)
-	// GrowBuffer grows a previous allocated buffer, reserving the appropriate memory space.
+	// GrowBuffer atomically grows a previous allocated buffer, reserving the appropriate memory space
+	// and releasing the old buffer. The copy parameter specifies the number of bytes to copy from
+	// the old buffer to the newly allocated buffer.
 	GrowBuffer(buf []byte, newsize, copy int) ([]byte, error)
 	// ReleaseBuffer releases a previous allocated buffer.
 	ReleaseBuffer(buf []byte)
@@ -52,17 +56,20 @@ type TransactionalScope interface {
 	Done()
 }
 
+// ServiceScope is the interface for service resource scopes
+type ServiceScope interface {
+	ResourceScope
+
+	// Name returns the name of this service
+	Name() string
+}
+
 // ProtocolScope is the interface for protocol resource scopes.
 type ProtocolScope interface {
 	ResourceScope
 
 	// Protocols returns the list of protocol IDs constrained by this scope.
 	Protocols() []protocol.ID
-	// IsDefault returns true if this is the default (global) scope.
-	IsDefault() bool
-
-	// AddStream adds a previously unassociated stream to this protocol scope.
-	AddStream(StreamScope) error
 }
 
 // PeerScope is the interface for peer resource scopes.
@@ -74,8 +81,11 @@ type PeerScope interface {
 
 	// OpenSconnect creates a new connection scope for this peer.
 	OpenConnection(dir Direction, usefd bool) (ConnectionScope, error)
-	// AddConnection adds a previously associated connection to this peer scope.
-	AddConnection(ConnectionScope) error
+
+	// OpenStream creates a new stream scope, with the specified protocols.
+	// An unnegotiated stream will have an empty protocol list and be initially unattached to any
+	// protocol scope.
+	OpenStream(dir Direction, proto ...protocol.ID) (StreamScope, error)
 }
 
 // ConnectionScope is the interface for connection resource scopes.
@@ -87,10 +97,8 @@ type ConnectionScope interface {
 	// It reeturns nil if the connection is not yet asociated with any peer.
 	PeerScope() PeerScope
 
-	// OpenStream creates a new stream scope, with the specified protocols.
-	// An unnegotiated stream will have an empty protocol list and be initially unattached to any
-	// protocol scope.
-	OpenStream(dir Direction, proto ...protocol.ID) (StreamScope, error)
+	// SetPeer sets the peer for a previously unassociated connection
+	SetPeer(peer.ID) error
 }
 
 // StreamScope is the interface for stream resource scopes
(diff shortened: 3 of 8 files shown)


network/rcmgr.go
```diff
@@ -0,0 +1,116 @@
...
+type ResourceScope interface {
+	// ReserveMemory reserves memory/buffer space in the scope.
+	ReserveMemory(size int) error
+	// ReleaseMemory explicitly releases memory previously reserved with ReserveMemory
+	ReleaseMemory(size int)
+
+	// GetBuffer reserves memory and allocates a buffer through the buffer pool.
+	GetBuffer(size int) ([]byte, error)
+	// GrowBuffer grows a previous allocated buffer, reserving the appropriate memory space.
+	GrowBuffer(buf []byte, newsize, copy int) ([]byte, error)
```
Stebalien, line 41: What's with the `copy` param?

network/rcmgr.go
```diff
@@ -0,0 +1,116 @@
...
+	ReleaseBuffer(buf []byte)
+
+	// Stat retrieves current resource usage for the scope.
+	Stat() ScopeStat
+}
+
+// TransactionalScope is a mixin interface for transactional scopes.
+type TransactionalScope interface {
+	// Done ends the transaction scope and releases associated resources.
+	Done()
```
Stebalien, line 52: Do we have any examples of how we think we might want to use these interfaces? We should start with some toy applications before we spend too much time on these APIs.

network/rcmgr.go
vyzo: to atomically copy the old buffer while growing/replacing it with a new one;  need to document this.

network/rcmgr.go
vyzo: The most basic hooks will be inside the transport and muxer implementations.
We need to cater to both them, at the bowels of the stack, and higher level users.

network/rcmgr.go
Stebalien: I'm still confused. Can't you just copy the old buffer anyways?

marten-seemann: Is the idea that stream muxers would use this?
I'm a bit worried about the API here, for example, this doesn't really match the quic-go API.

network/rcmgr.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// ProtocolScope is the interface for protocol resource scopes.
+type ProtocolScope interface {
+	ResourceScope
+
+	// Protocols returns the list of protocol IDs constrained by this scope.
+	Protocols() []protocol.ID
+	// IsDefault returns true if this is the default (global) scope.
+	IsDefault() bool
```
marten-seemann, line 62: What would be a non-default scope?

network/rcmgr.go
```diff
@@ -0,0 +1,116 @@
...
+
+// PeerScope is the interface for peer resource scopes.
+type PeerScope interface {
+	ResourceScope
+
+	// Peer returns the peer ID for this scope
+	Peer() peer.ID
+
+	// OpenSconnect creates a new connection scope for this peer.
+	OpenConnection(dir Direction, usefd bool) (ConnectionScope, error)
```
marten-seemann, line 76: This should probably be a `...ConnectionTrait`. WebRTC connections for example might be considered more expensive than a lean TCP / QUIC connection.

network/rcmgr.go
vyzo: I would like to make it atomically release and give you a new one. It really is a shortcut for Get new, Release old, without oversubscribing memory.

Maybe we dont need it, but of it is an old buffer growing by 1.5x it seems ugood to have it.

network/rcmgr.go
```diff
@@ -0,0 +1,116 @@
...
+	// AddConnection adds a previously associated connection to this peer scope.
+	AddConnection(ConnectionScope) error
```
marten-seemann, line 78: Why is this needed? A connection doesn't change it's peer.

network/rcmgr.go
```diff
@@ -0,0 +1,116 @@
...
+	ProtocolScope() ProtocolScope
+}
+
+// ScopeStat is a struct containing resource accounting information.
+type ScopeStat struct {
+	NumPeers   int
+	NumConns   int
+	NumStreams int
+
+	Memory int
```
marten-seemann, line 115:
```suggestion
	Memory int64
```

marten-seemann (commented):
Initial pass.

network/rcmgr.go
vyzo: I removed this method, it doesn't make much sense.

network/rcmgr.go
vyzo: maybe... but that would be reflected in their memory reservations, no?

Let's cross that bridge when we get there.

network/rcmgr.go
vyzo: removed this method, (unassociated) connections now have a method to set their peer once.

network/rcmgr.go
vyzo: done.

network/rcmgr.go
```diff
@@ -0,0 +1,136 @@
...
+	ErrResourceLimitExceeded = errors.New("resource limit exceeded")
+	ErrResourceScopeClosed   = errors.New("resource scope closed")
+)
+
+// ResourceManager is the interface to the network resource management subsystem
+type ResourceManager interface {
+	// GetSystem retrieves the system wide resource scope
+	GetSystem() ResourceScope
+	// GetDMZ retrieves the DMZ resource scope
+	GetDMZ() ResourceScope
```
marten-seemann, line 20: I still find it super weird to call this DMZ. Can we call it system scope?

network/rcmgr.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// PeerScope is the interface for peer resource scopes.
+type PeerScope interface {
+	ResourceScope
+
+	// Peer returns the peer ID for this scope
+	Peer() peer.ID
+
+	// OpenSconnect creates a new connection scope for this peer.
```
marten-seemann, line 87:
```suggestion
	// OpenConnection creates a new connection scope for this peer.
```

network/rcmgr.go
marten-seemann: ok, works for me

marten-seemann (commented):
Quick review today, more to follow tomorrow.

network/rcmgr.go
vyzo: Well, the system scope is the global hard limits; the DMZ is for floating resources that have not been associated with the appropriate scope yet (inbound connections, unassociated streams, etc).

But agreed, let's find a better name.

network/rcmgr.go
```diff
@@ -0,0 +1,136 @@
...
+// StreamScope is the interface for stream resource scopes
+type StreamScope interface {
+	ResourceScope
+	TransactionalScope
+
+	// ProtocolScope returns the protocol resource scope associated with this stream.
+	// It returns nil if the stream is not associated with any scope.
+	ProtocolScope() ProtocolScope
+	// SetProtocol sets the protocol for a previously unnegotiated stream
+	SetProtocol(proto protocol.ID) error
```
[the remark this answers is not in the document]
vyzo: Yes indeed. It also should error if you call it twice.

network/rcmgr.go
vyzo: how about we call it `FloatingScope`?

vyzo: or `TransientScope`

network/rcmgr.go
```diff
@@ -0,0 +1,136 @@
...
+
+// PeerScope is the interface for peer resource scopes.
+type PeerScope interface {
+	ResourceScope
+
+	// Peer returns the peer ID for this scope
+	Peer() peer.ID
+
+	// OpenConnection creates a new connection scope for this peer.
+	OpenConnection(dir Direction, useFD bool) (ConnectionScope, error)
```
marten-seemann, line 88: We've been talking about making it possible to dial a multiaddr without prior knowledge of the peer ID (we'd learn the peer ID from the handshake).

Should we simplify the API here and remove `OpenConnection` from the `PeerScope`? When dialing a connection to a known peer, we'd always go through the transient scope instead (and immediately call `SetPeer` then).

vyzo: makes sense, will do.

vyzo: done.

network/rcmgr.go
vyzo: renamed to `GetTransient`.

network/rcmgr.go
vyzo: removed the ncopy parameter, implementation should unconditionally copy.

network/rcmgr.go
```diff
@@ -0,0 +1,152 @@
...
+	ViewPeer(peer.ID, func(PeerScope) error) error
+
+	// OpenConnection creates a new connection scope not yet associated with any peer; the connection
+	// is scoped at the transient scope.
+	OpenConnection(dir Direction, usefd bool) (ConnectionScope, error)
+
+	// OpenStream creates a new stream scope, initially unnegotiated.
+	// An unnegotiated stream will be initially unattached to any protocol scope
+	// and constrained by the transient scope.
+	OpenStream(p peer.ID, dir Direction) (StreamScope, error)
```
marten-seemann, line 35: What was the reason this is not in the `PeerScope`?

vyzo: consistency, and it also simplifies locking.

network/rcmgr.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Stat retrieves current resource usage for the scope.
+	Stat() ScopeStat
+
+	// BeginTxn creates a new transactional scope rooted at this scope
+	BeginTxn() (TransactionalScope, error)
+}
+
+// Buffer is a wrapper for a buffer managed by a resource scope
+type Buffer interface {
```
marten-seemann, line 59: Should we make this a
```go
type Buffer []byte
```

I wish we could switch out the `append` function that Go uses!

vyzo: we can't, we can't hash or compare a slice, which makes it impossible to track  :(:(

network/rcmgr.go
```diff
@@ -0,0 +1,152 @@
...
+
+import (
+	"errors"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+)
+
+var (
+	ErrResourceLimitExceeded = errors.New("resource limit exceeded")
```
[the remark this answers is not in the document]
vyzo: they could I guess -- do we gain anything?

network/rcmgr.go
```diff
@@ -0,0 +1,152 @@
...
+	ReleaseMemory(size int)
+
+	// GetBuffer reserves memory and allocates a buffer through the buffer pool.
+	GetBuffer(size int) (Buffer, error)
+
+	// Stat retrieves current resource usage for the scope.
+	Stat() ScopeStat
+
+	// BeginTxn creates a new transactional scope rooted at this scope
+	BeginTxn() (TransactionalScope, error)
```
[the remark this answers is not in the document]
vyzo: ok

network/rcmgr.go
```diff
@@ -0,0 +1,152 @@
...
+	ServiceScope() ServiceScope
+	// SetService sets the service owning this stream
+	SetService(srv string) error
+
+	// PeerScope returns the peer resource scope associated with this stream.
+	PeerScope() PeerScope
+}
+
+// UserStreamScope is the user view of a StreamScope
+type UserStreamScope interface {
```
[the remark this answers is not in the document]
vyzo: Indeed, we could call the other two scopes `*ManagementScope`.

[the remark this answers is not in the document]
vyzo: In fact we should also narrow it to only include SetPeer, Stat, and BeginTxn.

network/rcmgr.go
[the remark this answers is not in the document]
vyzo: done.

network/rcmgr.go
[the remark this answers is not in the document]
vyzo: Renamed the full scopes to *ManagementScope and dropped the User prefix for user scopes. 

I haven't narrowed the interface though, that's something to think about.

network/rcmgr.go
vyzo: RIP buffers.

network/rcmgr.go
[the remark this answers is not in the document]
marten-seemann: Applications could catch this with a general "if it's just a temporary error, try again in a bit" logic.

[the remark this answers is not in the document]
vyzo: ok.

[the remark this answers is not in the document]
vyzo: done.

network/errors.go
```diff
@@ -1,6 +1,17 @@
 package network
 
-import "errors"
+import (
+	"errors"
+	"net"
+)
+
+type TemporaryError string
```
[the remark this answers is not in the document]
vyzo: sure, but isnt it generally useful?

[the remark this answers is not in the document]
vyzo: unexported for now; if anyone needs it, we can always upcase it later.

network/rcmgr.go
```diff
@@ -0,0 +1,132 @@
+package network
+
+import (
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+)
+
+// ResourceManager is the interface to the network resource management subsystem
+type ResourceManager interface {
```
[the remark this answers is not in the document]
vyzo: of course! We'll work on the docs once we have moved further forward with the integration and testing.

network/conn.go
```diff
@@ -68,3 +66,10 @@ type ConnStat interface {
 	// Stat stores metadata pertaining to this conn.
 	Stat() ConnStats
 }
+
+// ConnScope is the interface that one can mix into a connection interfac to give it a resource
```
marten-seemann, line 70: Typo: interface.

vyzo: fixed.

network/rcmgr.go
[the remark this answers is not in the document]
vyzo: updated.

network/rcmgr.go
```diff
@@ -0,0 +1,132 @@
...
+	PeerScope() PeerScope
+}
+
+// UserStreamScope is the user view of a StreamScope
+type StreamScope interface {
+	ResourceScope
+
+	// SetService sets the service owning this stream
+	SetService(srv string) error
+}
```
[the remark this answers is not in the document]
vyzo: The new big phat godoc on the ResourceManager should explain all this.

The Management scope distinction is also more salient now.

network/rcmgr.go
```diff
@@ -0,0 +1,132 @@
...
+	ReserveMemory(size int) error
+	// ReleaseMemory explicitly releases memory previously reserved with ReserveMemory
+	ReleaseMemory(size int)
+
+	// Stat retrieves current resource usage for the scope.
+	Stat() ScopeStat
+
+	// BeginTransaction creates a new transactional scope rooted at this scope
+	BeginTransaction() (TransactionalScope, error)
+}
```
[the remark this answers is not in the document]
vyzo: done.

network/rcmgr.go
```diff
@@ -0,0 +1,313 @@
...
+	// The caller owns the returned scope and is responsible for calling Done in order to signify
+	// the end of th scope's span.
+	OpenStream(p peer.ID, dir Direction) (StreamManagementScope, error)
+
+	// Close closes the resource manager
+	Close() error
+}
+
+// MemoryStatus is an indicator of the current level of available memory for scope reservations.
+type MemoryStatus int
```
marten-seemann, line 119:
```suggestion
type MemoryStatus uint8
```

vyzo: does this change anything? It is not stored anywhere, so it is not using memory and it is just a register in either case.

marten-seemann: Not sure, but it can't hurt, can it?

vyzo: yeah, I guess not and it might lead to shorter code in x86(-64) because it wont have the prefix byte!

network/rcmgr.go
```diff
@@ -0,0 +1,152 @@
...
+	// ReserveMemory reserves memory/buffer space in the scope.
+	ReserveMemory(size int) error
+	// ReleaseMemory explicitly releases memory previously reserved with ReserveMemory
+	ReleaseMemory(size int)
+
+	// GetBuffer reserves memory and allocates a buffer through the buffer pool.
+	GetBuffer(size int) (Buffer, error)
+
+	// Stat retrieves current resource usage for the scope.
+	Stat() ScopeStat
```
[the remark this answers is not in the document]
vyzo: yeah observability is very important.

We can add cli commands in ipfs and lotus to inspect resource usage.
So it's a keeper.

network/rcmgr.go
```diff
@@ -0,0 +1,317 @@
...
+// and file  descriptors. These account for both space and time used by
+// the stack, as each resource has a direct effect on the system
+// availability and performance.
+//
+// The modus operandi of the resource manager is to restrict resource usage at the time of
+// reservation. When a component of the stack needs to use a resource, it reserves it in the
+// appropriate scope. The resource manager gates the reservation against the scope applicable
+// limits; if the limit is exceeded, then an error (wrapping ErrResourceLimitExceeded) and it
+// is up the component to act accordingly. At the lower levels of the stack, this will normally
+// signal a filure of some sorts, like failing to opening a stream or a connection, which will
```
raulk, line 38:
```suggestion
// signal a failure of some sorts, like failing to opening a stream or a connection, which will
```

raulk, line 38: Did you consider defining a normalized errors for components to wrap when the resource manager denies an action, and this prevents progress?

network/rcmgr.go
```diff
@@ -0,0 +1,317 @@
...
+	ViewTransient(func(ResourceScope) error) error
+
+	// ViewService retrieves a service-specific scope.
+	ViewService(string, func(ServiceScope) error) error
+
+	// ViewProtocol views the resource management scope for a specific protocol.
+	ViewProtocol(protocol.ID, func(ProtocolScope) error) error
+
+	// ViewPeer views the resource management scope for a specific peer.
+	ViewPeer(peer.ID, func(PeerScope) error) error
```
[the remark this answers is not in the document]
vyzo: sure, can do.

network/rcmgr.go
vyzo: Yes, we have ErrResourceLimitExceeded which implements net.Error as temporary.

I should probably add something here about it.

network/rcmgr.go
```diff
@@ -0,0 +1,317 @@
...
+	ReserveMemory(size int, prio uint8) error
+
+	// ReleaseMemory explicitly releases memory previously reserved with ReserveMemory
+	ReleaseMemory(size int)
+
+	// Stat retrieves current resource usage for the scope.
+	Stat() ScopeStat
+
+	// BeginTransaction creates a new transactional scope rooted at this scope
+	BeginTransaction() (TransactionalScope, error)
```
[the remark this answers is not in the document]
vyzo: We currently error if the scope has been closed. But it's good to have room for growth too.

network/rcmgr.go
```diff
@@ -0,0 +1,317 @@
...
+	//
+	// The priority argument indicates the priority of the memory reservation. A reservation
+	// will fail if the available memory is less than (1+prio)/256 of the scope limit, providing
+	// a mechanism to gracefully handle optional reservations that might overload the system.
+	// For instance, a muxer growing a window buffer will use a low priority and only grow the buffer
+	// if there is no memory pressure in the system.
+	//
+	// The are 4 predefined priority levels, Low, Medium, High and Always,
+	// capturing common patterns, but the user is free to use any granularity applicable to his case.
+	ReserveMemory(size int, prio uint8) error
```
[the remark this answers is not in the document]
vyzo: sure, will do

network/rcmgr.go
```diff
@@ -0,0 +1,317 @@
...
+	// SetService sets the service owning this stream.
+	SetService(srv string) error
+}
+
+// ScopeStat is a struct containing resource accounting information.
+type ScopeStat struct {
+	NumStreamsInbound  int
+	NumStreamsOutbound int
+	NumConnsInbound    int
+	NumConnsOutbound   int
```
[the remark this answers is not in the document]
vyzo: what do you mean here? Relayed connections are just a virtual connection on top of some other connection and don't use fds.

transport/transport.go
```diff
@@ -29,6 +29,7 @@ type CapableConn interface {
 	mux.MuxedConn
 	network.ConnSecurity
 	network.ConnMultiaddrs
+	network.ConnScoper
```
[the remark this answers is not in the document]
vyzo: fixed.

network/rcmgr.go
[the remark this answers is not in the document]
vyzo: done.

network/rcmgr.go
```diff
@@ -0,0 +1,317 @@
...
+	// Reservation PriorityMedium is a reservation priority that indicates a reservation if the scope
+	// memory utilization is at 60% or less.
+	ReservationPriorityMedium uint8 = 152
+	// ReservationPriorityHigh is a reservation prioirity that indicates a reservation if the scope
+	// memory utilization is at 80% or less.
+	ReservationPriorityHigh uint8 = 203
+	// ReservationPriorityAlways is a reservation priority that indicates a reservation if there is
+	// enough memory, regardless of scope utilization.
+	ReservationPriorityAlways uint8 = 255
+)
```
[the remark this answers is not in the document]
vyzo: They are just predefined convenience constats; user is free to ignore and just use a raw number.

network/rcmgr.go
```diff
@@ -0,0 +1,317 @@
...
+	// The transient scope accounts for resources that are in the process of
+	// full establishment.  For instance, a new connection prior to the
+	// handshake does not belong to any peer, but it still needs to be
+	// constrained as this opens an avenue for attacks in transient resource
+	// usage. Similarly, a stream that has not negotiated a protocol yet is
+	// constrained by the transient scope.
+	ViewTransient(func(ResourceScope) error) error
+
+	// ViewService retrieves a service-specific scope.
+	ViewService(string, func(ServiceScope) error) error
```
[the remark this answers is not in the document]
vyzo: No, the service does not deal with conns at all, it only deals with streams.
Conns are really shared resources, so there isn't much of a point to have them owned by a service.

network/rcmgr.go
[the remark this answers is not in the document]
vyzo: We do count them however in stat.

network/rcmgr.go
vyzo: wait, there is a reference to ErrResourceLimitExceeded already here; it is defined in errors.go

network/rcmgr.go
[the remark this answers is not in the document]
vyzo: done.

marten-seemann (approved):

PR #230, opened 2021-12-19
remove the DialTimeout

marten-seemann:
The timeout should be a configuration option for the swarm, not a global variable.

Related change in the swarm: https://github.com/libp2p/go-libp2p-swarm/pull/302.

We should probably do the same for the `AcceptTimeout`.

### DIFF
--- a/transport/transport.go
+++ b/transport/transport.go
@@ -14,11 +14,6 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
-// DialTimeout is the maximum duration a Dial is allowed to take.
-// This includes the time between dialing the raw network connection,
-// protocol selection as well the handshake, if applicable.
-var DialTimeout = 15 * time.Second
-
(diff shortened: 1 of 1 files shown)


vyzo (commented):
wait, dont transports themselves use that?

marten-seemann:
> wait, dont transports themselves use that?

Nope.

TCP uses an option: https://github.com/libp2p/go-tcp-transport/blob/1c17791c35777091afcb56a0707503bf82532a62/tcp.go#L98-L103
QUIC just uses the quic-go defaults.

marten-seemann:
Confirmed by building go-ipfs using this PR and https://github.com/libp2p/go-libp2p-swarm/pull/302.

vyzo (commented):
well ok, @Stebalien was this ever useful?

Stebalien (approved):
I think we ended up using custom timeouts in transports, because the actual timeout is transport dependent.

vyzo (approved):

PR #232, opened 2022-01-02
introduce a transport.Upgrader interface

marten-seemann:


vyzo (commented):
do we need the deprecated methods?

marten-seemann:
I think the circuit code is still using them, but we could clean that up in the next release. Let's remove them.

vyzo:
yeah

vyzo (approved):

PR #238, opened 2022-03-14
chore: add `String()` method to `IDSlice` type

EclesioMeloJunior:
### Description

- Initialize and preallocate the slice of strings
- use the method `String` for each `ID` and put in the slice 
- return a string of ids separated by `, `

Tiny demonstration: https://go.dev/play/p/kd_nd-edWnr

Original suggestion -> https://github.com/ChainSafe/gossamer/pull/2267#discussion_r826306828


marten-seemann (commented):
Why is this needed? The linked issue doesn't seem to have a lot to do with it. The link is about a race condition, certainly a stringer function can't help with that.

marten-seemann (approved):

PR #240, opened 2022-03-30
fix: switch to go-multicodec mappings

lidel:
The mappings in go-cid were maintained by hand and are invalid.
More details in https://github.com/ipfs/go-cid/pull/137

This PR is switching to go-multicodec which has correct mappings that are generated from https://github.com/multiformats/multicodec/blob/master/table.csv, removing the surface for human error.

### DIFF
--- a/peer/peer.go
+++ b/peer/peer.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ipfs/go-cid"
 	ic "github.com/libp2p/go-libp2p-core/crypto"
 	b58 "github.com/mr-tron/base58/base58"
+	mc "github.com/multiformats/go-multicodec"
 	mh "github.com/multiformats/go-multihash"
 )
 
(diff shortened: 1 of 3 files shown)


peer/peer.go
```diff
@@ -162,13 +163,9 @@ func Encode(id ID) string {
...
-	ty := c.Type()
-	if ty != cid.Libp2pKey {
-		s := cid.CodecToStr[ty]
-		if s == "" {
-			s = fmt.Sprintf("[unknown multicodec %d]", ty)
-		}
-		return "", fmt.Errorf("can't convert CID of type %s to a peer ID", s)
+	code := mc.Code(c.Type())
+	if code != mc.Libp2pKey {
+		return "", fmt.Errorf("can't convert CID of type %s to a peer ID", code.String())
```
mvdan, line 168:
```suggestion
		return "", fmt.Errorf("can't convert CID of type %q to a peer ID", code)
```

fmt's %s will already call the String method by default :) plus, using %q is a bit better so we don't print something confusing if e.g. a code string has a space.

mvdan (approved):
SGTM

PR #243, opened 2022-04-14
feat: harden encoding/decoding functions against panics

Stebalien:
Part of https://github.com/libp2p/go-libp2p/issues/1389

These kinds of functions:

1. Handle user input.
2. Often have out-of-bounds, null pointer, etc bugs.
3. Have completely isolated logic where local panics are unlikely to cause memory corruption elsewhere.

### DIFF
--- a/internal/catch/catch.go
+++ b/internal/catch/catch.go
@@ -1,0 +1,18 @@
+package catch
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"runtime/debug"
+)
+
+var panicWriter io.Writer = os.Stderr
+
+// HandlePanic handles and logs panics.
+func HandlePanic(rerr interface{}, err *error, where string) {
+	if rerr != nil {
+		fmt.Fprintf(panicWriter, "caught panic: %s\n%s\n", rerr, debug.Stack())
+		*err = fmt.Errorf("panic in %s: %s", where, rerr)
+	}
+}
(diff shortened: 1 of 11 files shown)


marten-seemann (commented):
As discussed:
* let's keep this to crypto operations
* add stderr output

marten-seemann:
@Stebalien Can you update this PR to log the stack trace to stderr?

internal/catch/catch.go
```diff
@@ -0,0 +1,14 @@
+package catch
+
+import (
+	"fmt"
+	"os"
+	"runtime/debug"
+)
+
+func CatchAndLog(err *error, where string) {
```
marten-seemann, line 9: I like the idea, but I don't think `recover` can be moved to a separate function. I added a random `panic` to one of the functions and it didn't work.

marten-seemann (changes requested):

internal/catch/catch.go
Stebalien: This works:

```go
package main

import (
	"fmt"
)

func myRecover() {
	if err := recover(); err != nil {
		fmt.Println("reccovered: ", err)
	}
}

func foo() {
	defer myRecover()

	panic("foo")
}

func main() {
	foo()
}
```


I'll see if I can figure out what I did wrong...

Stebalien: I've added a test case.

marten-seemann (approved):

PR #250, opened 2022-05-20
remove OpenedStream and ClosedStream from Notifiee interface

marten-seemann:
As far as I can tell, nobody is using the `OpenedStream` and `ClosedStream` notification.

This is because they're not really useful: You can get a notification that the swarm opened a new stream, but you can read, write, close or reset the stream without interfering with the application. Therefore you won't even know which libp2p protocol is being negotiated on that stream.


vyzo (approved):
fine by me.

@Stebalien anyone using this?

Stebalien:
We ahve some tests inside bitswap, but that's it. It looks like elrond may have been using this for some form of connection management, but really the hook isn't all that useful.

I'd remove it and figure out a way to fix upstream.

MarcoPolo (approved):

PR #251, opened 2022-05-24
release v0.16.0

marten-seemann:


vyzo (approved):
changelog? We should document the notifee change if anything else.

marten-seemann:
> changelog? We should document the notifee change if anything else.

Will do, in the release notes.

PR #258, opened 2022-06-15
Add canonical log for misbehaving peers

MarcoPolo:
Add canonical logging for misbehaving peers. See https://github.com/libp2p/go-libp2p/issues/1598 for more details.

Why go-libp2p-core? I wish I could put this in the go-libp2p mono repo, but I don't want to require go-libp2p just so that the resource manager can use this. But I don't have strong opinions on where this lives. It is unfortunate that this causes a lot of extra stuff to be added to go.mod/sum

I do have strong opinions on consolidating this repo into `go-libp2p`, we should. Managing multiple changes across go-libp2p-core, go-libp2p, and resource manager has been like walking through a tar pit. More discussion here https://github.com/libp2p/go-libp2p/issues/1556.


canonicallog/canonicallog.go
```diff
@@ -0,0 +1,50 @@
...
+// user to easily identify these nodes across protocols and libp2p.
+func LogMisbehavingPeer(p peer.ID, peerAddr multiaddr.Multiaddr, err error, msg string) {
+	log.Errorf("CANONICAL_MISBEHAVING_PEER: peer=%s addr=%s err=%v msg=%s", p, peerAddr.String(), err, msg)
+}
+
+// LogMisbehavingPeer is the canonical way to log a misbehaving peer.
+// Protocols should use this to identify a misbehaving peer to allow the end
+// user to easily identify these nodes across protocols and libp2p.
+func LogMisbehavingPeerNetAddr(p peer.ID, peerAddr net.Addr, originalErr error, msg string) {
+	ipStrandPort := strings.Split(peerAddr.String(), ":")
```
marten-seemann, line 25: Any reason not to use manet.FromNetAddr?

marten-seemann (changes requested):
Should we add a “component” string, so that we know where the error occurred? For example, this would be the application protocol, or TLS or Noise.

canonicallog/canonicallog.go
MarcoPolo: oh nice! Thanks

MarcoPolo:
> Should we add a “component” string, so that we know where the error occurred? For example, this would be the application protocol, or TLS or Noise.

Yeah I think this makes sense, but I want to highlight that this doesn't need to be super specific unless we notice that it's useful to differentiate if the TLS handshake or Noise handshake failed. Otherwise we can start at component=handshake.

canonicallog/canonicallog.go
```diff
@@ -0,0 +1,31 @@
...
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+var log = logging.WithSkip(logging.Logger("canonical-log"), 1)
+
+// LogMisbehavingPeer is the canonical way to log a misbehaving peer.
+// Protocols should use this to identify a misbehaving peer to allow the end
+// user to easily identify these nodes across protocols and libp2p.
+func LogMisbehavingPeer(p peer.ID, peerAddr multiaddr.Multiaddr, component string, err error, msg string) {
+	log.Errorf("CANONICAL_MISBEHAVING_PEER: peer=%s addr=%s component=%s err=%v msg=%s", p, peerAddr.String(), component, err, msg)
```
marten-seemann, line 18: No strong opinion, but should we use `Errorfw` instead? Might make it easier for tools to parse.

MarcoPolo: You mean `Errorw`? In my experience parsing `key=value` pairs are much easier than parsing json. But not a strong opinion here.

marten-seemann (approved):

PR #271, opened 2022-06-30
Release v0.19.0

MarcoPolo:
Releases https://github.com/libp2p/go-libp2p-core/pull/257

This is purposely done in its own release so that we can get the earlier changes usable before this change which relies on https://github.com/libp2p/go-libp2p-resource-manager/pull/47 to be merged.

😢 all these repos...


marten-seemann (approved):
My hope is that v0.20.0 will be the repo deprecation release!

Issue #252, opened 2022-05-24
btcec update leads to `go mod tidy` failure


marten-seemann:
We just released v0.16.0, which include the btcec update (https://github.com/libp2p/go-libp2p-core/pull/245).

`go mod tidy` now produces the following error (see https://github.com/libp2p/go-libp2p/pull/1521):

```
❯ go mod tidy
github.com/libp2p/go-libp2p imports
        github.com/libp2p/go-libp2p-core/crypto imports
        github.com/btcsuite/btcd/btcec/v2/ecdsa tested by
        github.com/btcsuite/btcd/btcec/v2/ecdsa.test imports
        github.com/btcsuite/btcd/chaincfg/chainhash loaded from github.com/btcsuite/btcd@v0.22.0-beta,
        but go 1.16 would fail to locate it:
        ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules:
        github.com/btcsuite/btcd v0.22.0-beta (/Users/marten/src/go/pkg/mod/github.com/btcsuite/btcd@v0.22.0-beta/chaincfg/chainhash)
        github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0 (/Users/marten/src/go/pkg/mod/github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.0)

To proceed despite packages unresolved in go 1.16:
        go mod tidy -e
If reproducibility with go 1.16 is not needed:
        go mod tidy -compat=1.17
For other options, see:
        https://golang.org/doc/modules/pruning
```

Updating btced to v2.2.0 (from v2.1.3) doesn't solve the problem. I also haven't succeeded in manually adding `github.com/btcsuite/btcd/chaincfg/chainhash` to go-libp2p's `go.mod` file.
People had a similar issue in https://github.com/btcsuite/btcd/issues/1839, but I couldn't make it work in go-libp2p either.

This is now blocking the go-libp2p v0.20.0 release. I'm planning to revert #245, unless we can get this solved within the next 24 hours or so.

cc @brianathere

brianathere:
Looking at this now.

vyzo:
Can we fix our CI globally to do 1.17 compat tidy?
We've had this problem elsewhere too.

marten-seemann:
> Can we fix our CI globally to do 1.17 compat tidy?

We could, but the problem is that _every_ user of go-libp2p would not be able to run a normal `go mod tidy` any more.

[...]

PR #272, opened 2022-07-05
Remove btcsuite/btcd dep

MarcoPolo:
I don't think need to pull in this btcd dependency. btcec is now it's own module.

This will also prevent us from forcing the btcd dep upgrade on downstream users. It breaks somethings especially https://github.com/whyrusleeping/ledger-filecoin-go which is used by lotus.


marten-seemann:
Will we run into the problem fixed in https://github.com/libp2p/go-libp2p-core/pull/254 again?

My conclusion back then was that this will have to wait until we stop using Go 1.17.

MarcoPolo:
go checks pass, so maybe not?

I think removing this dep would side step that problem by not having btcd to conflict with.

MarcoPolo:
Was able to sucessfully remove this dep from go-libp2p as well which unblocked lotus from being able to update: https://github.com/filecoin-project/lotus/pull/8970

marten-seemann:
> go checks pass, so maybe not?

The checks passed in this repo, but not when updating go-libp2p to the new version. I'm ok with merging this (actually, I'd be very happy to get rid of this hack!!), but we should make sure that go-libp2p will still pass CI when we make this change.
More details in https://github.com/libp2p/go-libp2p-core/issues/252.

MarcoPolo:
> The checks passed in this repo, but not when updating go-libp2p to the new version.

Seems like the tests are passing in go-libp2p too: https://github.com/libp2p/go-libp2p/pull/1641

marten-seemann (approved):
Then let’s get rid of this hack!

PR #275, opened 2022-07-07
deprecate peer.Encode in favor of peer.ID.String

marten-seemann:
Trying to reduce the function sprawl in this package a bit.


Stebalien (approved):
Seems reasonable.

PR #280, opened 2022-08-18
crypto/secp256k1: Remove btcsuite intermediary.

### LINKED ISSUE #11: Remove dependency on btcsuite
We currently depend on that for the infrequently used secp256k1 key type. Unfortunately, we _do_ have users using this, at least for IPNS. We may be able to deal with this by supporting out-of-tree plugins for new key types.

We could also try extracting the package we need from upstream.

davecgh:
This updates the `crypto/secp256k1` code to make use of the `dcrec/secp256k1/v4` module directly instead of using `btcec/v2` which itself is now just a shim around `dcrec/secp256k1/v4` anyway.

This has the benefit of removing the additional `github.com/btcsuite/btcd/chaincfg/chainhash` dependency since `dcrec/secp256k1/v4` is its own module and does not rely on it.

It also updates to the latest `v4.1.0` release which implements direct key generation and has some other nice optimizations that speed up signature verification as compared to the `v4.0.1` release.

Closes #11.


Kubuxu (approved):
In theory, this is a breaking change but it is one I would be ok with.

davecgh:
> In theory, this is a breaking change but it is one I would be ok with.

I was a little bit concerned about that as well due to changing the type aliases, but I confirmed that it actually is not because the types in `btcec/v2` are themselves just aliases to the same underlying type this changes them too.  For example, see the [btcec/v2.PrivateKey](https://github.com/btcsuite/btcd/blob/0f49e1000633b0af6d7bbef97d69671f4c99e37e/btcec/privkey.go#L14) type declaration which is an alias for `dcrec/secp256k1/v4.PrivateKey`.  The same is true for [btcec/v2.PublicKey](https://github.com/btcsuite/btcd/blob/0f49e1000633b0af6d7bbef97d69671f4c99e37e/btcec/pubkey.go#L40). 

In order to check that is indeed the case, I intentionally kept the type assertion in `KeyPairFromStdKey` as `*btcec.PrivateKey` and ran the `TestKeyPairFromKey` test which generates the key with `dcrec.GeneratePrivateKey` (which is actually a `*dcrec.PrivateKey`) and then calls `KeyPairFromStdKey` with it to ensure they are considered identical by the compiler.

Here is the diff relative to this PR to be more concrete about what I just explained:

```diff
diff --git a/crypto/key_not_openssl.go b/crypto/key_not_openssl.go
index 8e63d9f..28792f8 100644
--- a/crypto/key_not_openssl.go
+++ b/crypto/key_not_openssl.go
@@ -9,7 +9,7 @@ import (
        "crypto/ed25519"
        "crypto/rsa"

-       dcrec "github.com/decred/dcrd/dcrec/secp256k1/v4"
+       btcec "github.com/btcsuite/btcd/btcec/v2"
 )

 // KeyPairFromStdKey wraps standard library (and secp256k1) private keys in libp2p/go-libp2p-core/crypto keys
@@ -30,7 +30,7 @@ func KeyPairFromStdKey(priv crypto.PrivateKey) (PrivKey, PubKey, error) {
                pub, _ := pubIfc.(ed25519.PublicKey)
                return &Ed25519PrivateKey{*p}, &Ed25519PublicKey{pub}, nil

-       case *dcrec.PrivateKey:
+       case *btcec.PrivateKey:
                sPriv := Secp256k1PrivateKey(*p)
                sPub := Secp256k1PublicKey(*p.PubKey())
                return &sPriv, &sPub, nil
diff --git a/go.mod b/go.mod
index 630296b..5f80bd6 100644
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.17

 require (
        github.com/benbjohnson/clock v1.3.0
+       github.com/btcsuite/btcd/btcec/v2 v2.1.3
        github.com/decred/dcrd/dcrec/secp256k1/v4 v4.1.0
        github.com/gogo/protobuf v1.3.1
        github.com/ipfs/go-cid v0.2.0
```

```bash
$ go test -run=TestKeyPairFromKey
PASS
ok      github.com/libp2p/go-libp2p-core/crypto 0.474s
```

marten-seemann:
We've just moved go-libp2p-ocre into go-libp2p/core: https://github.com/libp2p/go-libp2p/pull/1683.

Can you please recreate this PR in the go-libp2p repo?

marten-seemann:
Really sorry for this confusion! I realize this is far from ideal.
We've been trying to minimize the transition time, we'll be deprecating this repo this week, but we first have to cut the new go-libp2p release and forward all the types (for backwards compatibility).

davecgh:
> Can you please recreate this PR in the go-libp2p repo?

Done.  libp2p/go-libp2p#1689.

PR #282, opened 2022-09-06
feat: forward crypto/pb

Jorropo:


marten-seemann (approved):
Are the maps not needed?

https://github.com/libp2p/go-libp2p-core/blob/acfb38a4ec7a5902cee8997e085ac65fef4d7b63/crypto/pb/crypto.pb.go#L35-L47

Jorropo:
@marten-seemann the maps are forwarded aswell:

https://github.com/libp2p/go-libp2p-core/pull/282/files#diff-81f634294a542589b62215ad6612b1185d0d93a5b87d924a37f5876c982ee508R13-R16

Also let's not fix the CI, this is just staticheck complaining that I am using deprecated code.

